PR #16, opened 2024-05-07
Use float-abi=hard on armv7hf

apljungquist:
Without it linking to some libraries fail at run-time with an obscure error message.

- [ ] Document the error message
- [ ] Document how to reproduce

### DIFF
--- a/Cross.toml
+++ b/Cross.toml
@@ -23,5 +23,5 @@ passthrough = [
     "PKG_CONFIG_PATH_thumbv7neon_unknown_linux_gnueabihf=/opt/axis/acapsdk/sysroots/armv7hf/usr/lib/pkgconfig:/opt/axis/acapsdk/sysroots/armv7hf/usr/share/pkgconfig",
     "PKG_CONFIG_LIBDIR_thumbv7neon_unknown_linux_gnueabihf=/opt/axis/acapsdk/sysroots/armv7hf/usr/lib/pkgconfig:/opt/axis/acapsdk/sysroots/armv7hf/usr/share/pkgconfig",
     "CARGO_TARGET_THUMBV7NEON_UNKNOWN_LINUX_GNUEABIHF_LINKER=arm-linux-gnueabihf-gcc",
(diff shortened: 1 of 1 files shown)


Cross.toml
```diff
@@ -23,5 +23,5 @@ passthrough = [
     "PKG_CONFIG_PATH_thumbv7neon_unknown_linux_gnueabihf=/opt/axis/acapsdk/sysroots/armv7hf/usr/lib/pkgconfig:/opt/axis/acapsdk/sysroots/armv7hf/usr/share/pkgconfig",
     "PKG_CONFIG_LIBDIR_thumbv7neon_unknown_linux_gnueabihf=/opt/axis/acapsdk/sysroots/armv7hf/usr/lib/pkgconfig:/opt/axis/acapsdk/sysroots/armv7hf/usr/share/pkgconfig",
     "CARGO_TARGET_THUMBV7NEON_UNKNOWN_LINUX_GNUEABIHF_LINKER=arm-linux-gnueabihf-gcc",
-    "CARGO_TARGET_THUMBV7NEON_UNKNOWN_LINUX_GNUEABIHF_RUSTFLAGS=-C link-args=--sysroot=/opt/axis/acapsdk/sysroots/armv7hf",
+    "CARGO_TARGET_THUMBV7NEON_UNKNOWN_LINUX_GNUEABIHF_RUSTFLAGS=-C link-args=--sysroot=/opt/axis/acapsdk/sysroots/armv7hf -C link_arg=-mfloat-abi=hard",
```
guoxe, line 26: Is `-C link-arg=` needed given that we supply `-C link-args=` initially? Wondering if it could be something like:
`-C link-args=--sysroot=/opt/axis/acapsdk/sysroots/armv7hf -mfloat-abi=hard`

apljungquist:
Half a year later this still bug has not reared its head, so closing this until someone stumbles across it again or takes time to reproduce it.

For posterity I think it presented when using mdb on armv7hf and/or artpec7.

PR #70, opened 2024-08-14
Add explicit platform to Dockerfile FROM statements

### LINKED ISSUE #51: Dev container does not launch on ARM mac
The dev container fails to launch with the error `[qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory]` on an ARM-based mac.

You can add `--platform=linux/amd64` to the `FROM` statements to overcome this, but seems kinda hacky.

juho:
### Describe your changes

Added explicit platform statements to the Dockerfile

- Fixes #51


juho:
I don't know if this is the best solution, but it's the one that worked for me. `runArgs` did not seem to do anything for some reason.

apljungquist (approved):

PR #89, opened 2024-09-04
fix(cargo-acap-sdk)!: Use dev profile for install command

apljungquist:
The intended benefit is to speed up subsequent uses of the run command by building the same cache.

I find myself using the install command primarily to:
- perform the initial installation after checking out an app or re- initializing a device.
- perform a full installation because the manifest has changed.

In both cases I am likely to then use the run command, so any space savings on the device are uninteresting.

One remaining benefit is faster installation time, but for the uses outlined above the time savings from a shared cache are greater.

Before: 75
- install: 43 (28 build + 1 pack + 14 upload)
- run: 32 (21 build + 5 pack + 7 upload)

After: 70
- install: 59 (22 build + 5 pack + 31 upload)
- run: 11 (0 build + 5 pack + 7 upload)

(vapix_access built on i7-6700K and  installed on P8815-2)

### DIFF
--- a/crates/acap-vapix/src/apis/applications_upload.rs
+++ b/crates/acap-vapix/src/apis/applications_upload.rs
@@ -123,13 +123,12 @@ impl<'a> UploadRequest<'a> {
             .post(PATH)
             .map_err(HttpRpcError::ParseUrl)?
             .replace_with(|b| {
-                dbg!(b
-                    .header(
-                        CONTENT_TYPE,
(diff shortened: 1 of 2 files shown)


apljungquist:
With a warm release cache the roles are likely reversed, making this seem like a deterioration. Maybe worth revisiting if the cost of packing the app can be reduced. Until then, a more likely path to speedup seems to be making the size of the release binary smaller.

PR #96, opened 2024-10-07
Improve mdb

kjughx:
1. It doesn't make much sense to bundle OwnedOrBorrowedError when they serve different use cases, better split them into separate types.
2. Message borrows *const mdb_sys::mdb_message_t so should contain a marker with a lifetime bound.
3. The callbacks used in mdb are hidden behind pointers to pointers to DSTs when they could just be generic pointers. This also simplifies the UX since a consumer doesn't have to wrap callbacks in Box::new().

### DIFF
--- a/apps/consume_analytics_metadata/src/main.rs
+++ b/apps/consume_analytics_metadata/src/main.rs
@@ -1,7 +1,7 @@
 use std::{ffi::CStr, process::abort, thread::sleep, time::Duration};
 
 use log::{error, info};
-use mdb::{Connection, Subscriber, SubscriberConfig};
+use mdb::{error::BorrowedError, Connection, Message, Subscriber, SubscriberConfig};
 
 const TOPIC: &CStr = c"com.axis.analytics_scene_description.v0.beta";
 const SOURCE: &CStr = c"1";
@@ -9,34 +9,31 @@ const SOURCE: &CStr = c"1";
 fn main() {
     acap_logging::init_logger();
 
-    let connection = Connection::try_new(Some(Box::new(|e| {
+    let connection = Connection::try_new(Some(|e: BorrowedError| {
         error!("Not connected because {e:?}");
         abort();
-    })))
+    }))
     .unwrap();
 
     let config = SubscriberConfig::try_new(
         TOPIC,
         SOURCE,
-        Box::new(|message| {
+        |message: Message| {
             let payload = String::from_utf8_lossy(message.payload());
             let libc::timespec{ tv_sec, tv_nsec } = message.timestamp();
             info!("message received from topic: {TOPIC:?} on source: {SOURCE:?}: Monotonic time - {tv_sec}.{tv_nsec:0>9}. Data - {payload}");
-        }),
+        },
     )
     .unwrap();
-    let _subscriber = Subscriber::try_new(
-        &connection,
-        config,
-        Box::new(|e| match e {
+    let _subscriber =
+        Subscriber::try_new(&connection, config, |e: Option<BorrowedError>| match e {
             None => info!("Subscribed"),
             Some(e) => {
                 error!("Not subscribed because {e:?}");
                 abort();
             }
-        }),
-    )
-    .unwrap();
+        })
+        .unwrap();
 
     loop {
         sleep(Duration::from_secs(1));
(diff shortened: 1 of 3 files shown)


crates/mdb/src/error.rs
```diff
@@ -0,0 +1,100 @@
...
+use std::fmt::{Debug, Display, Formatter};
+use std::marker::PhantomData;
+
+unsafe fn pchar_to_string(p_value: *const c_char) -> String {
+    assert!(!p_value.is_null());
+    let value = String::from(CStr::from_ptr(p_value).to_str().unwrap());
+    value
+}
+
+pub trait MdbError<Ptr> {
```
apljungquist, line 11: What are some benefits of defining a trait for this?

In particular, do we really want `new` and `as_ref` to be available to users of the library?

crates/mdb/src/lib.rs
```diff
@@ -21,124 +20,64 @@ macro_rules! suppress_unwind {
...
-    Borrowed(*const mdb_sys::mdb_error_t),
-}
-pub struct Error(OwnedOrBorrowedError);
-
-impl Debug for Error {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Code: {}; Message: {:?};", self.code(), self.message())
-    }
-}
+pub trait Callback<T>: FnMut(T) + Send + 'static {}
```
apljungquist, line 23: Interesting, I have not seen this pattern for handling callbacks before. What are some benefits of using our own trait instead of relying on only the `Fn*` traits?

Have you seen any other crates or authors using it (i.e. is there anywhere I can learn more about it)?

apps/consume_analytics_metadata/src/main.rs
apljungquist, line 1: Do the tests pass when you run them e.g. with `asdk test`?

I'm using an M416-V with 11.11.73

crates/mdb/src/lib.rs
```diff
@@ -218,80 +160,75 @@ impl Drop for SubscriberConfig {
...
                 on_done as *mut c_void,
                 &mut error,
-            );
-            match (ptr.is_null(), error.is_null()) {
-                (false, false) => {
-                    panic!("mdb_subscriber_create_async returned both a connection and an error")
-                }
-                (false, true) => {
-                    let on_message = config.on_message;
-                    config.on_message = std::ptr::null_mut();
```
apljungquist, line 261: If we don't store the `on_message` callback on the `Subscriber`, will it not be dropped as soon as this `try_new` returns?

crates/mdb/src/lib.rs
```diff
@@ -300,25 +237,27 @@ impl<'a> Drop for Subscriber<'a> {
...
             mdb_sys::mdb_subscriber_destroy(&mut self.ptr);
-            drop(Box::from_raw(self.on_done));
-            drop(Box::from_raw(self.on_message));
+            drop(Box::from_raw(self.on_done))
         }
     }
 }
 
-unsafe impl<'a> Send for Subscriber<'a> {}
+unsafe impl Send for Subscriber<'_> {}
```
apljungquist, line 245: Nice touch.

crates/mdb/src/lib.rs
```diff
@@ -21,124 +20,64 @@ macro_rules! suppress_unwind {
...
 
 impl Connection {
     // TODO: Consider adopting a builder-like pattern.
-    pub fn try_new(on_error: Option<Box<OnError>>) -> Result<Self, Error> {
+    pub fn try_new<F>(on_error: Option<F>) -> Result<Self, OwnedError>
+    where
+        F: for<'a> Callback<BorrowedError<'a>>,
+    {
         debug!("Creating {}...", any::type_name::<Self>());
-        unsafe {
```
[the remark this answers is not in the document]
kjughx: I dropped that commit.

crates/mdb/src/error.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl Drop for OwnedError {
+    fn drop(&mut self) {
+        unsafe {
+            mdb_sys::mdb_error_destroy(&mut self.ptr as *mut *mut _);
+        }
+    }
+}
+unsafe impl Send for OwnedError {}
```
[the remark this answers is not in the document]
kjughx: Yes I think pointless, BorrowedError is only used in callbacks which exist in some context in mdb, it doesn't make sense to send a BorrowedError somewhere else since it will probably be dropped by the callback caller once the callback returns.

crates/mdb/src/lib.rs
kjughx: Good point. Could Subscriber just own SubscriberConfig which then is dropped when Subscriber is dropped?

apps/consume_analytics_metadata/src/main.rs
kjughx: They do now, I forgot to update the code in the test.

crates/mdb/src/lib.rs
apljungquist: It could. I actually wrote it like that at first to minimize the complexity, but then I changed my mind thinking that it is unnecessary to tie up the `mdb_subscriber_config_t` for the lifetime of the connection. Are there reasons other than complexity for going back? Is there anything in particular that you think contributes to complexity with the current solution (assuming complexity is the reason you want to change this)?

crates/mdb/src/error.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn message(&self) -> String;
+}
+
+pub struct OwnedError {
+    ptr: *mut mdb_sys::mdb_error_t,
+}
+
+pub struct BorrowedError<'a> {
+    ptr: *const mdb_sys::mdb_error_t,
+    _marker: PhantomData<&'a mdb_sys::mdb_error_t>,
```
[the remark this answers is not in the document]
apljungquist: As I think more about this, I start doubting that `BorrowedError` should have an associated lifetime; maybe it is better/correct that the API only gives out `&'a BorrowedError` instead, where `'a` is the lifetime of the reference. This looks at least superficially similar to `&'a CStr`.

I want to do some research on this as well; If you are impatient about getting the new callbacks merged, I would suggest scaling back the scope of the PR to only those.

crates/mdb/src/lib.rs
```diff
@@ -21,87 +20,37 @@ macro_rules! suppress_unwind {
...
 pub struct Connection {
     ptr: *mut mdb_sys::mdb_connection_t,
-    on_error: *mut Box<OnError>,
+    on_error: *mut OnError,
 }
 
 impl Connection {
     // TODO: Consider adopting a builder-like pattern.
-    pub fn try_new(on_error: Option<Box<OnError>>) -> Result<Self, Error> {
+    /// Passing `None` as the `@on_error` callback requires qualifying the generic like so:
```
apljungquist, line 37: As far as I know it is convention in Rust to reference parameters without an `@` prefix, see for instance `Box::new`.

apljungquist:
Consider splitting the PR, the parts that are purely a refactoring are an especially good candidate. That way we can get something merged faster by focusing our attention on a smaller change. It may also help us write a good commit message; I'm trying to get into the habit of following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), but since I'm inconsistent at best this is not a requirement.

crates/mdb/src/error.rs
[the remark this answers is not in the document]
kjughx: See #116, there the API only exposes `Error` and `&Error`. The owned `Error` as a return value and `&Error` in the callback.

apljungquist:
@kjughx does this PR contain anything not covered by #110, #115 and #116?

kjughx:
> @kjughx does this PR contain anything not covered by...

No, I'll close it.

PR #102, opened 2024-10-16
chore: Track size of apps

apljungquist:
Because this is an important property and we want to make sure that the size actually decreases when we make improvements and that it does not inadvertently increase.

The total size of all apps can be reported like:
```console
$ cat apps-aarch64.filesize | cut -f1 | paste -sd+ | bc
66644
```

Using `--profile app` would make the size smaller and, arguably, more representative but it would also require us to change what we build so it is skipped for now:
```console
$ cat apps-aarch64.filesize | cut -f1 | paste -sd+ | bc
4676
```

`Makefile`:
- Don't use the `--total` option to reduce the risk of merge conflicts.
- Don't use the `--human-readable` option to make it easier to sum the numbers afterwards.

### DIFF
--- a/apps-aarch64.filesize
+++ b/apps-aarch64.filesize
@@ -1,0 +1,10 @@
+5808	target/acap/bounding_box_example_0_0_0_aarch64.eap
+5676	target/acap/consume_analytics_metadata_0_0_0_aarch64.eap
+860	target/acap/embedded_web_page_0_0_0_aarch64.eap
+5596	target/acap/hello_world_0_0_0_aarch64.eap
+5676	target/acap/inspect_env_0_0_0_aarch64.eap
+5648	target/acap/licensekey_handler_0_0_0_aarch64.eap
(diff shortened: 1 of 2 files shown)


apps-aarch64.filesize
```diff
@@ -0,0 +1,10 @@
+5808	target/acap/bounding_box_example_0_0_0_aarch64.eap
```
guoxe, line 1: Are these sizes in KB?

apljungquist: KiB I think:

> Display values are in units of the first available SIZE from --block-size, and the DU_BLOCK_SIZE, BLOCK_SIZE and BLOCKSIZE environment variables.  Otherwise,  units  default  to  1024
       bytes (or 512 if POSIXLY_CORRECT is set).

guoxe: I see, thanks!

guoxe (approved):
lgtm

PR #108, opened 2024-10-18
feat: Add edge storage wrapper and example

apljungquist:
This wrapper omits ergonomic wrappers and leans hard into making the API flexible and familiar to experienced ACAP developers. Intended benefits include:
- Referencing the functions looks nice and familiar; `axevevent_list` becomes `axevent::list`, etc.
- Fewer API design decisions have to be considered speeding up the development of new bindings.
- Allows contributors specializing in writing unsafe code and FFI bindings to focus on this demanding task while allowing contributors who are interested in API design can work on ergonomic crates without worrying about accidentally violating any safety preconditions.

If I remember correctly one big reason why I started collocating the flexible and ergonomic APIs was that I needed to access the flexible API internals when developing a replica of the event system for running unit tests on host; I no longer intend to do this.

With the new example app the sum of file sizes increase:
```console
$ cat apps-aarch64.filesize | cut -f1 | paste -sd+ | bc
41768
```

`apps/axstorage_example/src/main.rs`:
- Use thread locals instead of `Mutex` and `AtomicU32` to emphasize that the application is single threaded.

`bin/samba*`:
- Add utilities for easily hosting a network storage that devices can connect to for testing the example app.

`crates/axstorage-sys/src/lib.rs`:
- Make type aliases public because I find using the aliases makes comparing the code I write to the C documentation and examples easier.

`crates/axstorage/src/lib.rs`:
- Using `GStringPtr` instead of a custom type proved difficult when converting from raw pointers in callback trampolines. It also does not implement `Send` event though I suspect this may be possible. I nonetheless chose to use it because maybe there is a reason for it not being `Send` and working around this in the example app was not difficult, I get many trait implementations for free and it generally seems like a good idea to lean into GLib when it already features in the C API.
- Use a custom subscription type despite making the Rust API less similar to the C API because we need ensure that the callback is freed and that it is freed at an appropriate time.
- Exclude the unknown type because the docs state that the function "storage type or UNKNOWN_TYPE if failure" and that "then a location for a GError is also provided" which I interpret to mean that this type will be returned if and only if an error is set, and in that case we return an error anyway so the user would never see the unknown type.


crates/axstorage/src/flex.rs
```diff
@@ -0,0 +1,376 @@
...
+    // TODO: SAFETY
+    unsafe {
+        let storage_type = try_func!(ax_storage_get_type, storage.raw);
+        debug_assert_ne!(storage_type, AXStorageType_UNKNOWN_TYPE);
+        Ok(Type::from_raw(storage_type))
+    }
+}
+
+// TODO: Verify safety of passing the result of this to C
+fn to_mut_ptr(s: &mut GStringPtr) -> *mut gchar {
```
guoxe, line 374: I think it's worth noting that this assumes that GStringPtr is implemented as a transparent wrapper around gchar and will have unintended results if that is not the case.

PR #110, opened 2024-10-23
feat(mdb): Add marker for Message

kjughx:
Message borrows *const mdb_sys::mdb_message_t and should have a marker to indicate that.


apljungquist:
Can you point to any concrete reason why the current implementation is unsafe, makes for an awkward interface or is otherwise inferior to the proposed implementation?

I'm inclined to keep a single `Message` type because I think it will be conceptually simpler when the C library gives us an owned message
(I hope I will be able to convince the authors to do so because there seems to be no technical limitations preventing this and having to clone it before sending it across threads is suboptimal).

I did some research and I think there are broadly two patterns that we can employ here;
Since both are represented in `glib` I will use those names to refer to the pattern:

- `Borrowed`: the current implementation (but since we never have an owned message we can skip the `Drop` implementation and, consequently, the `ManuallyDrop`) 
- `BorrowedObject`: the proposed implementation

Pros of `Borrowed`:

- Used by at least one reputable library [^4].
- Single type is easier to understand, document, etc.
- Less code assuming we will eventually handle both a borrowed and an owned message.
- We benefit from blanket implementations for `&T`.
  - This can be mitigated by implementing `Deref<Target=&OwnedMessage> for BorrowedMessage`. 

Pros of `BorrowedObject`:

- Used by the standard library [^3] and at least one more reputable library [^1].
- Textbook example of when to use `PhantomData<&'a T>` [^2].

Notice how all the benefits of `BorrowedObject` are essentially arguments from authority;
I don't understand why this pattern is beneficial but smarter people than me nonetheless decided to use it.
The only reason I have found so far is that it makes FFI easier [^7]:

> BorrowedFd looks like it'll be nicer than &OwnedFd because it can use repr(transparent) and participate in FFI directly. This means users can avoid touching Raw* values entirely.

A third pattern that seems related, but less so, is slices;
Types such as `CStr`, `GStr`, and `Path` are based on slices.
These are similar to `Borrowed` in that they are not generic over some lifetime and are typically found only as references.
However, they are also similar to `BorrowedObject` in that they have an owned counterpart and can be owned e.g. `Box<str>`[^5].

p.s. apologies for the late and possibly incoherent reply; I nerd sniped myself with some make stuff but still want to give you a response this weekend.

[^1]: https://gtk-rs.org/gtk-rs-core/stable/0.19/docs/glib/object/struct.BorrowedObject.html
[^2]: https://doc.rust-lang.org/nomicon/phantom-data.html
[^3]: https://doc.rust-lang.org/beta/std/os/fd/struct.BorrowedFd.html
[^4]: https://gtk-rs.org/gtk-rs-core/stable/0.19/docs/glib/translate/struct.Borrowed.html
[^5]: https://doc.rust-lang.org/src/alloc/string.rs.html#1956 
[^6]: https://rust-lang.github.io/rfcs/3128-io-safety.html#ownedfd-and-borrowedfdfd
[^7]: https://github.com/rust-lang/rfcs/pull/3128#issuecomment-850926113

kjughx:
> Can you point to any concrete reason why the current implementation is unsafe, makes for an awkward interface or is otherwise inferior to the proposed implementation?

I don't think the current implementation actually is unsafe as it is, since `Message` is passed by reference to the callback. However, consider this:

```diff
- type OnMessage = dyn FnMut(&Message) + Send + 'static;
+ type OnMessage = dyn FnMut(Message) + Send + 'static;
+ unsafe impl Send for Message {}
+ usnafe impl Sync for Message {}
```
i.e. it takes `Message` by value and not reference. Inside `Message` however is a reference.

Without `PhantomData`, the following is legal:

```rust
let connection = Connection::try_new(None).unwrap();
let (tx, rx) = std::sync::mpsc::channel();

let config = SubscriberConfig::try_new(
    TOPIC,
    SOURCE,
    Box::new(move |message| {
        tx.send(message).unwrap();
    }),
)
```
which rustc happily accepts. Problem is: `message` is deallocated at the end of the callback and `rx` will receive a dangling pointer.

With `PhantomData`:
```shell
2 |     let (tx, rx) = std::sync::mpsc::channel();
   |          -- `tx` declared here, outside of the closure body
...
7 |         Box::new(move |message| {
   |                        ------- `message` is a reference that is only valid in the closure body
8 |             tx.send(message).unwrap();
   |             ^^^^^^^^^^^^^^^^ `message` escapes the closure body here
```

Granted this isn't what the current implementation does but it's a pitfall that the compiler would catch.

crates/mdb/src/lib.rs
apljungquist, line 1: Should we change the callback too?

Now that the lifetime is on the message, the lifetime of the reference is not needed to prevent it from escaping.

apljungquist:
> Granted this isn't what the current implementation does but it's a pitfall that the compiler would catch.

So the objective is making it easier for us as library writers to avoid mistakes. I can get onboard with that :smile:.

crates/mdb/src/lib.rs
kjughx: I don't have a strong opinion, I'll go with whatever you think is best :)

apljungquist:
Nice. We'll probably need to merge main and update the checksum as well.

apljungquist (approved):

PR #115, opened 2024-10-26
feat(mdb): Improve ergonomics for consumer

kjughx:
It's very clunky for the consumer to have to pass a Box<_> callback. The functions that accept callbacks should be generic over the callback. The ownership of the callbacks was also unclear.

`crates/mdb/src/lib.rs`:
- Make the functions that accept callbacks generic over the callback. This improves ergonomics for the consumer because they don't have to wrap the callback in a Box.
- Add a new type `Deferred` which owns the callback pointer that mdb-sys borrows. The type implements Drop which means we don't have to manually drop the callback when it's not used anymore or if there was an error.

`apps/consume_analytics_metadata/src/main.rs`:
- Make the necessary changes to accommodate the changes in crates/mdb, specifically remove the Box around the callbacks.

This depends on #110.


apljungquist:
Nice. Let's update the checksums and merge this.

apljungquist (approved):

apljungquist:
I'll merge this next time I'm at my computer

PR #116, opened 2024-10-26
feat(mdb): Improve Error

kjughx:
There are two use-cases of Error:
 - Owned: When mdb-sys functions fail they return an owned Error.
 - Borrowed: The callbacks passed to mdb-sys functions accept a reference to an Error from mdb-sys.

There are now two error types: Error (public to consumer) and BorrowedError (public to the crate). The owned error returned by mdb-sys is transparently an Error.

The references to Error are wrapped by BorrowedError within the crate and exposed as &Error to consumers. This is possible because both types wrap a pointer to mdb_sys::mdb_error so BorrowedError can be coerced into an Error.

`crates/mdb/src/error.rs`:
- Break out all error related definitions and implementations.
- Add new (public to crate) type `BorrowedError` which represents a reference to an mdb error.
- Consumers will only interact with `Error` and `&Error` for more cohesiveness.

`crates/mdb/src/lib.rs`:
- The callbacks now accept references. `&Message<'_>` instead of `Message<'_>` and `&Error` as opposed to an "owned `Error(OwnedOrBorrowedError::Borrowed`".

This change also contains the changes from #110 and #115. Not sure if this is the right way to approach to a "stack of PRs". Let me know if I should remove those changes and provide only the diff between #115 and this.


crates/mdb/src/error.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn message(&self) -> String {
+        unsafe { pchar_to_string(self.as_ref().message) }
+    }
+}
+
+impl std::ops::Deref for BorrowedError<'_> {
+    type Target = Error;
+    // SAFETY: self.ptr is effectively a reference to Error provided by mdb-sys for the duration
+    // of a callback. BorrowedError is tied to the lifetime of the contained pointer and since
+    // Error does not implement Copy there is no way to leak the contained pointer.
```
apljungquist, line 65: One precondition that we should document here once it is guaranteed is that the pointer is not null because I'm pretty sure casting a null pointer to a reference is UB

crates/mdb/src/lib.rs
```diff
@@ -21,123 +20,68 @@ macro_rules! suppress_unwind {
...
 pub struct Connection {
     ptr: *mut mdb_sys::mdb_connection_t,
-    on_error: *mut Box<OnError>,
+    _on_error: Option<Deferred>,
 }
 
 impl Connection {
     // TODO: Consider adopting a builder-like pattern.
-    pub fn try_new(on_error: Option<Box<OnError>>) -> Result<Self, Error> {
+    /// Passing `None` as the `@on_error` callback requires qualifying the generic like so:
```
apljungquist, line 30: Please remove `@` in front of the argument name because it would be inconsistent with how the rest of this project refers to arguments.

crates/mdb/src/lib.rs
```diff
@@ -151,61 +95,75 @@ impl Drop for Connection {
...
     }
 }
 
 unsafe impl Send for Connection {}
 unsafe impl Sync for Connection {}
 
+struct Deferred(Option<Box<dyn FnOnce()>>);
+impl Drop for Deferred {
+    fn drop(&mut self) {
+        assert!(self.0.is_some());
```
apljungquist, line 108: What's the benefit of the `assert` considering the `unwrap` on the next line?

crates/mdb/src/lib.rs
```diff
@@ -151,61 +95,75 @@ impl Drop for Connection {
...
+        self.0.take().unwrap()()
```
apljungquist, line 109: Don't we need to call the function here?

`new` takes a pointer and creates a function that, when called, creates a box then drops that box. Now, if we never call the function, then the box is never created and never dropped.

apljungquist (commented):
I'm holding off on reviewing more of the code until I better understand the implications of implementing `Deref` and converting `BorrowedError` to `&OwnedError` more generally.

crates/mdb/src/lib.rs
```diff
@@ -132,7 +133,7 @@ pub struct SubscriberConfig {
 impl SubscriberConfig {
     pub fn try_new<F>(topic: &CStr, source: &CStr, on_message: F) -> Result<Self, Error>
     where
-        F: for<'a> FnMut(Message<'a>) + Send + 'static,
+        F: for<'a> FnMut(&Message<'a>) + Send + 'static,
```
apljungquist, line 136: Should this change?

crates/mdb/src/lib.rs
```diff
@@ -250,17 +251,17 @@ impl<'a> Subscriber<'a> {
 
     unsafe extern "C" fn on_done<F>(error: *const mdb_sys::mdb_error_t, user_data: *mut c_void)
     where
-        F: FnMut(Option<Error>) + Send + 'static,
+        F: FnMut(Option<&Error>) + Send + 'static,
```
apljungquist, line 254: Nice.

apps/consume_analytics_metadata/src/main.rs
```diff
@@ -4,15 +4,15 @@
...
+use mdb::{Connection, Error, Subscriber, SubscriberConfig};
 
 const TOPIC: &CStr = c"com.axis.analytics_scene_description.v0.beta";
 const SOURCE: &CStr = c"1";
 
 fn main() {
     acap_logging::init_logger();
 
-    let connection = Connection::try_new(Some(|e| {
+    let connection = Connection::try_new(Some(|e: &Error| {
```
apljungquist, line 15: When I test I see this is necessary but I don't understand why, so if you know please share.

crates/mdb/src/error.rs
```diff
@@ -32,37 +32,22 @@ impl Display for Error {
...
-                }
-                OwnedOrBorrowedError::Borrowed(_) => {}
-            }
+            mdb_sys::mdb_error_destroy(&mut self.ptr);
         }
     }
 }
 unsafe impl Send for Error {}
-// Note that error is not Sync
+// Note that Error is not Sync
```
apljungquist, line 40: Good catch.

crates/mdb/src/error.rs
```diff
@@ -73,3 +58,22 @@ impl Error {
         unsafe { pchar_to_string(self.as_ref().message) }
     }
 }
+
+impl std::ops::Deref for BorrowedError<'_> {
+    type Target = Error;
+    // SAFETY: self.ptr is effectively a reference to Error provided by mdb-sys for the duration
```
apljungquist, line 64: > self.ptr is effectively a reference to Error

This does not sound right to me. Given that

```rust
#[repr(transparent)]
struct `Error(*mut mdb_error_t)
```

I think `*const mdb_error_t` is more like `Error` than `&Error`. Am I missing something?

Can we back up the theory with a model implementation or mock implementation that we can run under miri?

I experimented a bit with modelling high risk patterns here: https://github.com/AxisCommunications/acap-rs/pull/140

crates/mdb/src/error.rs
```diff
@@ -73,3 +58,22 @@ impl Error {
...
+    // of a callback. BorrowedError is tied to the lifetime of the contained pointer and since
+    // Error does not implement Copy there is no way to leak the contained pointer.
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*(self.ptr as *const Error) }
+    }
+}
+
+impl BorrowedError<'_> {
+    pub(crate) fn new(ptr: *const mdb_sys::mdb_error_t) -> Self {
```
apljungquist, line 73: I think this should be unsafe because there are preconditions that the caller must ensure for this function to be sound. I know it kept the safe marking from `new_borrowed`, so you don't have to change it now. But we should change it eventually.

crates/mdb/src/lib.rs
kjughx: I wasn't sure, I thought for cohesiveness it'd be nice if "callbacks are passed shared references" even though `Message` already is, at least in essence, a reference. It does look a little weird that the lifetime is on `Message` and not `&` though. Ideally it would be a `&'a Message`, but maybe that's something to consider in a different PR, I'll remove it here.

crates/mdb/src/error.rs
kjughx: You are right, this was not correct. It should look like this:

```Rust
        unsafe { &*((&self.ptr as *const *const sys::error_t) as *const Error) }
```
That is to say, `self.ptr` is effectively a "const" `Error`, whatever that means. In this context though, since `Error` is transparently a `mdb_error_t`, we want to treat `&self.ptr` as a `&Error`. 

I validated this through miri, see the new commit.

apps/consume_analytics_metadata/src/main.rs
kjughx: I don't exactly know either, the compiler thinks the closure might capture the environment which limits the lifetime on `e` which must by any lifetime. Making it non-optional works though, is there a particular reason this callback is optional but all the other ones arent?

crates/mdb/src/error.rs
apljungquist: Yikes. But it looks correct as long as `self.ptr` does not move, which it doesn't.

apps/consume_analytics_metadata/src/main.rs
apljungquist: I think the underlying C API works that way. If I were to guess why I'd say it is plausible that we want to create a connection without a callback telling us how it went, but it makes no sense to subscribe to messages without a method of handling the received messages.

crates/mdb/src/error.rs
```diff
@@ -1,21 +1,21 @@
...
-enum OwnedOrBorrowedError {
-    Owned(*mut mdb_sys::mdb_error_t),
-    Borrowed(*const mdb_sys::mdb_error_t),
+pub struct Error {
+    ptr: *mut mdb_sys::mdb_error_t,
+}
+
+pub(crate) struct BorrowedError<'a> {
+    ptr: *const mdb_sys::mdb_error_t,
+    _marker: PhantomData<&'a mdb_sys::mdb_error_t>,
```
apljungquist, line 17: Do you happen to know if `mdb_sys::mdb_error_t` is necessary or if `()` would suffice since we are only trying to capture the lifetime?

apljungquist (approved):
Nice. I assume the unit tests pass as well?

crates/mdb/src/error.rs
kjughx: I'm not actually sure, but I think being verbose here is nice :)

apps/consume_analytics_metadata/src/main.rs
kjughx: Fair enough

PR #120, opened 2024-10-29
Add Larod

JsGjKJzi:
The acap-rs workspace doesn't currently provide wrappers for the [larod library](https://axiscommunications.github.io/acap-documentation/docs/api/src/api/larod/html/index.html). When it is ready, I hope this merge request will provide that.

## Immediate goals:

- Provide safe Rust wrappers around the Larod functions, and abstract into rust structs and impls where appropriate.
- Provide tests to verify functionality.

Since larod uses peripheral hardware, I plan to build the tests and execute them on an Axis P1468-LE.

- [ ] I have performed a self-review of my own code
- [ ] I have verified that the code builds perfectly fine on my local system

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,12 +56,12 @@ bbox-sys = { path = "crates/bbox-sys" }
 cargo-acap-build = { path = "crates/cargo-acap-build" }
 cli-version = { path = "crates/cli-version" }
 device-manager = { path = "crates/device-manager" }
+larod = { path = "crates/larod" }
+larod-sys = { path = "crates/larod-sys" }
 licensekey = { path = "crates/licensekey" }
 licensekey-sys = { path = "crates/licensekey-sys" }
 mdb = { path = "crates/mdb" }
 mdb-sys = { path = "crates/mdb-sys" }
-larod = { path = "crates/larod" }
-larod-sys = { path = "crates/larod-sys" }
 
 [workspace.package]
 edition = "2021"


JsGjKJzi:
Just adding a comment that I am still relatively new to Rust, so suggestions to improve my code is very welcome. I will also be learning about functionality of the larod library during the process of creating this wrapper. So, guidance on interfacing with larod is also very welcome.

JsGjKJzi:
Looking at the larodError documentation a bit more closely, I realize what I have so far won't work, and maybe larod cannot be wrapped?

https://axiscommunications.github.io/acap-documentation/docs/api/src/api/larod/html/structlarodError.html

Notably larodError, which is passed to nearly every function, must be uninitialized. It might be impossible for rust to produce such a pointer https://doc.rust-lang.org/std/mem/union.MaybeUninit.html.

apljungquist:
Since you say the code is not ready I will focus on answering your questions and assist you when you ask for it for now. Consequently I haven't looked much at the code but I address your comments below.

---

> Just adding a comment that I am still relatively new to Rust

Diving in at the deep end with unsafe. A word of warning; In my opinion unsafe combines the worst of C(++) and Rust, and writing safe wrappers is especially frustrating since many of the safety aspects that we need to guarantee are not documented - sometimes not even considered -  by the library authors. To make matters worse, we often cannot even test empirically if our code is correct because UB does not mean that it will segfault or do anything weird right now, it means it could do it now or at any point in the future. So, know know what you are getting yourself into, that it is ok to quit and don't let this experience discourage you from using Rust because there's another side to the language that is carefree and fun :sun_with_face: :rainbow: 

What is your programming background? I will try to adjust my communication accordingly.

> I will also be learning about functionality of the larod library during the process of creating this wrapper.

That makes two of us.

> Notably larodError, which is passed to nearly every function, must be uninitialized.

I have found that sometimes the documentation should be taken with a grain of salt (I have found errors) and other times it is important to read it carefully. In any case the examples can be a good supporting resource. in `vdo_larod.c` [^1] we find this:

```c
larodError* error = NULL;
// ...
if (!larodConnect(&conn, &error)) {
```

We clearly don't pass an uninitialized variable but a pointer to null. I think this is what the documentation means when it says "An uninitialized **handle** to an error".

Compare this to `send_event.c` [^2] which is makes heavy use of glib and, presumably, follows glib error conventions:

```c
GError* error = NULL;
// ...
if (!ax_event_handler_declare(event_handler,
                                  // ...
                                  &error)) {
```

[^1]: https://github.com/AxisCommunications/acap-native-sdk-examples/blob/9b00b2fdf23672f8910421653706572201c2ed8b/vdo-larod/app/vdo_larod.c#L85
[^2]: https://github.com/AxisCommunications/acap-native-sdk-examples/blob/9b00b2fdf23672f8910421653706572201c2ed8b/axevent/send_event/app/send_event.c#L170

JsGjKJzi:
Thanks for the feedback. My background is varied engineering (R&D, so jack of all trades master of none sort of thing). I've written production C++ for embedded microcontrollers, C++ for AI using OpenCV and libtorch, lots of Python, and a few web backend type services using Rust. Again, I'm no expert, but I've done a lot of dabbling in various things.

I'm ok diving into the deep end. Just more to learn hopefully :).

One piece from the larod docs that I worry a bit about is this quote, "error can also be NULL if one does not want any error information." I worry that if I initialize the pointer to Null, we'll just never get any error information back.

apljungquist:
I'm comparing to axevent again because I recently worked on those bindings.

I think this is what passing a null pointer would look like: https://github.com/AxisCommunications/acap-native-sdk-examples/blob/9b00b2fdf23672f8910421653706572201c2ed8b/axevent/send_event/app/send_event.c#L167

Compare that to passing a (not-null) pointer to a null pointer: https://github.com/AxisCommunications/acap-native-sdk-examples/blob/9b00b2fdf23672f8910421653706572201c2ed8b/axevent/send_event/app/send_event.c#L176

JsGjKJzi:
@apljungquist I just set up the P1468-LE camera and set up a remote testing workflow. I've added some documentation and configuration files in [this commit](https://github.com/AxisCommunications/acap-rs/pull/120/commits/c05a0a1625c7669a5060ea5712dc5dbfe588083f).

It might be useful for you in your endeavors. I have confirmed it works for the larod library so far on the P1468-LE running OS 11.11.73. It might be worth considering as a cherry pick, or I could open a separate merge request for it.

apljungquist:
Neat. I  didn’t know about `runner = ["/workspaces/acap-rs/remote-test.sh"]` so I implemented this in `cargo-asap-sdk`. I’m traveling over the weekend but when I’m back I’ll try out your solution and think about how I would like to leverage it.

I currently think of tests in two categories:

- Library test. This is what you have written and implemented support. We typically upload them to `/tmp/` and run them as root. Starting with AXIS OS 12.0 this will be difficult or impossible to do 
- Application tests. These can be packaged as an EAP and installed on the device as any app. They run as the same user as the app would, which is required for APIs like message broker. This likely will continue to work or, in the worst case, the test output would have to be retrieved from the system logs.

Where you aware of `cargo-acap-rs test`? Asking because if not then I have a marketing problem.

remote-test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -e
+if [[ -z "${CARGO_TEST_CAMERA}" ]]; then
```
apljungquist: What is the purpose of checking this?

remote-test.sh
```diff
@@ -0,0 +1,10 @@
...
+    f=`basename $1`
+    scp "$1" $CARGO_TEST_CAMERA:.
+    # echo $f
+    ssh $CARGO_TEST_CAMERA "chmod +x /root/$f" 
+    ssh $CARGO_TEST_CAMERA "/root/$f"
+    # ^ note: may need to change this line, see https://stackoverflow.com/q/9379400
```
apljungquist, line 9: When I try the exit code of the shell script is 0 when the tests pass and 101 (which seems to be the code cargo expected) when they fail. What behavior do you observe or worry about?

apljungquist (commented):
I like the simplicity of this script but before I make any suggestions about how to upstream it I have a couple of questions to make sure I understand it.

remote-test.sh
JsGjKJzi: This overall would probably need more work. If a runner is defined, i.e. `target.aarch64-unknown-linux-gnu.runner` executables are always passed to the runner instead of executed directly. I was trying to create an escape hatch so that if someone doesn't have a camera, and hasn't set `CARGO_TEST_CAMERA`, it would still attempt to execute directly like normal. However, I hadn't really finished the script.

remote-test.sh
JsGjKJzi: I see the same behavior. The stack overflow link was provided as a comment on return values that I kept in case something didn't work out. It seems to apply more to Android, which was part of the inspiration, and doesn't seem to affect cargo.

JsGjKJzi:
> Were you aware of `cargo-acap-sdk test`? Asking because if not then I have a marketing problem.
I might have been, but have been caught up in thinking through the API I'd like to achieve that I may have forgot about it. I have definitely used `cargo-acap-sdk build` to build an eap.

Using that may ultimately be a better approach than using the runner as discussed above.

apljungquist:
Happy to see you are still working on this 🙂 What's are you aiming for (the whole API vs some subset, a minimal safe abstraction vs highly ergonomic abstraction, integration with VDO, etc) and how far do you think you have come?
Is there anything I can do to help?

I'm sorry I still haven't made time to study the APIs and your implementation.

JsGjKJzi:
Yep, I'm still working on it when I find time. I was initially planning on doing the whole API, but I'm starting to realize some elements of the larod API don't make a ton of sense. I'm also thinking covering the whole API is a bit more work than I initially realized. So, I'm leaning more towards a minimal but functional library for now that strikes a balance between being a relatively thin wrapper and providing an idiomatic Rust API. 

I am also starting to work on the VDO library so I can get frames from the camera. That might be a bit more difficult looking at the [imageprovider.c](https://github.com/AxisCommunications/acap-native-sdk-examples/blob/main/vdo-larod/app/imgprovider.c) file in the vdo-larod example. Setting up background threads and interfacing the GLib will take a bit of thinking through. I'm aiming for an API that would work something like this:
```Rust
fn main() -> Result<()> {
    let stream = Stream::builder()
        .channel()
        .format()
        .width()
        .height()
        .build();
    stream.start();
    while let Ok(frame) = stream.next_frame() {
        ...
    }
    Ok(())
}
```

### Example liblarod Confusion
The [larodCreateTensors](https://axiscommunications.github.io/acap-documentation/3.5/api/larod/html/larod_8h.html#aededa9e269d87d0f1b7636a007760cb2) function exists and simply takes a `size_t` number of tensors to create. I believe the library allocates some vector-like data structure on the heap and hands back a pointer to that data structure. The docs say to call [larodDestroyTensors](https://axiscommunications.github.io/acap-documentation/3.5/api/larod/html/larod_8h.html#afac99dfef68ffe3d513008aaac354ae0) which takes a pointer to the `larodConnection`. 

These two functions have me a bit confused. the `larodCreateTensors` does not seem to require a `larodConnection`, so it would seem the lifetime is like `'static`. However, since the data structure is a C-allocated thing, we have to hand it back to C to deallocate and we don't seem to have a good function for doing that. Other functions, such as [larodCreateModelInputs](https://axiscommunications.github.io/acap-documentation/3.5/api/larod/html/larod_8h.html#adefd8c496e10eddced5be85d93aceb13) take a pointer to the `larodModel`, for which a `larodConnection` is needed in the first place. So the lifetime of these tensors is inherently tied to the lifetime of the `larodConnection`. Fine. Also, when [larodDisconnect](https://axiscommunications.github.io/acap-documentation/3.5/api/larod/html/larod_8h.html#ab8f97b4b4d15798384ca25f32ca77bba) is called, any allocated tensors can be deallocated since they can easily be associated with that initial `larodConnection`.

I did a quick test (I believe, I can't find it now) where I used `larodCreateTensors`,  `larodConnect`, and `larodDestroyTensors(larodConnection* conn, ... )` and got no errors. That seems to invite the possibility for any client to deallocate any other client's tensors as long as they have a pointer to the container of tensors, which doesn't seem particularly safe. Maybe there is a use case for that, but I thought I would leave out a Rust version of `larodCreateTensors` for now.

### Code Conventions
One thing I was struggling with was the combination of supported [preprocessor configurations](https://axiscommunications.github.io/acap-documentation/3.5/api/larod/html/md__opt_builder-doc_larod_doc_preprocessing.html). We know a-priori what backends support what kind of operations. So, I would really like to create a Rust Preprocessor API which encodes these supported combinations in the type system. We could then lean on the Rust type system to make it mostly impossible to code an incompatible combination of parameters. I couldn't find a nice way of doing it with Enums. The [typestate pattern](https://docs.rust-embedded.org/book/static-guarantees/typestate-programming.html) might be able to do it but looked like it was going to quite a bit of library code to do. 

For now, I'm following the C library pretty closely with a builder pattern and doing the appropriate checks in a `.build()` method at runtime and returning an `Err` if an invalid combination is requested.

JsGjKJzi:
@apljungquist I'd like your opinion on something. I'm looking at the various [supported backends for larod](https://axiscommunications.github.io/acap-documentation/docs/api/src/api/larod/html/md__opt_builder-doc_larod_doc_nn-inference.html). My initial thought was that since we know what combinations of hardware (e.g. ARTPEC 8) and model type (e.g. tflite) are, I could use Rust's type system to emit compile time errors if invalid combinations are defined.

Here's an example:
```rust

mod inference {
    pub trait PrivateSupportedBackend {}
}

pub trait SupportedBackend: inference::PrivateSupportedBackend {
    fn new() -> Self;
    fn as_str(&self) -> &str;
}

// Model types
pub struct TFLite;
pub struct CVFlowNN;
pub struct Native;

// Hardware
pub struct CPU;
pub struct Artpec8DLPU;


impl inference::PrivateSupportedBackend for InferenceBackend<TFLite, CPU> {}
impl inference::PrivateSupportedBackend for InferenceBackend<TFLite, Artpec8DLPU> {}

pub struct InferenceBackend<M, H> {
    mode: M,
    hardware: H,
}

impl SupportedBackend for InferenceBackend<TFLite, CPU> {
    fn new() -> InferenceBackend<TFLite, CPU> {
        InferenceBackend {
            mode: TFLite,
            hardware: CPU,
        }
    }

    fn as_str(&self) -> &str {
        "cpu-tflite"
    }
}

impl SupportedBackend for InferenceBackend<TFLite, Artpec8DLPU> {
    fn new() -> InferenceBackend<TFLite, Artpec8DLPU> {
        InferenceBackend {
            mode: TFLite,
            hardware: Artpec8DLPU,
        }
    }
    fn as_str(&self) -> &str {
        "axis-a8-dlpu-tflite"
    }
}

fn main() {}
   // let a = larod::InferenceBackend::<Native, ArmNNCPU>::new(); // Compiler error
  let b = larod::InferenceBackend::<TFLite, Artpec8DLPU>::new(); // Fine, since Artpec8DLPU can support TFLite models.
}
```

This works, but I wonder if it is the right approach. An alternative is to just define some enums and verify things at runtime:

```rust
#[derive(Debug)]
pub enum Device {
    CPU,
    ARTPEC7GPU,
    ARTPEC8DLPU,
    ARTPEC9DLPU,
}

#[derive(Debug)]
pub enum Backend {
    TFLite(Device),
}

impl Backend {
    pub fn as_str(&self) -> &str {
        match self {
            Backend::TFLite(d) => match d {
                Device::CPU => "cpu-tflite",
                Device::ARTPEC7GPU => "axis-a7-gpu-tflite",
                Device::ARTPEC8DLPU => "axis-a8-dlpu-tflite",
                Device::ARTPEC9DLPU => "a9-dlpu-tflite",
            },
        }
    }
}
```

This approach has the benefit of more ergonomic matching and possibly deserialization from `Session.devices`:
```rust
let session = Session::new();
    let devices = session
        .devices()
        .context("Couldn't list available devices")?;
```
```bash
[2025-01-15T16:07:11Z INFO  test] Devices:
[2025-01-15T16:07:11Z INFO  test] cpu-tflite (0)
[2025-01-15T16:07:11Z INFO  test] axis-ace-proc (0)
[2025-01-15T16:07:11Z INFO  test] cpu-proc (0)
[2025-01-15T16:07:11Z INFO  test] axis-a8-dlpu-tflite (0)
[2025-01-15T16:07:11Z INFO  test] axis-a8-dlpu-native (0)
[2025-01-15T16:07:11Z INFO  test] axis-a8-dlpu-proc (0)
[2025-01-15T16:07:11Z INFO  test] axis-a8-gpu-proc (0)
```

apljungquist:
I've been happy to follow along with your conversations with Daniel and to see that you are still making progress.

I agree that it would be nice to leverage Rust's type system to avoid runtime errors. I've only had time to take a quick look at the documentation so far, I'll try to make more time this weekend if needed. Here are my thoughts so far:

1. Are device and backend the same thing? ('The device name of this backend is "cpu-tflite"; a device handle can be retrieved by providing this device name to `larodGetDevice()`' makes it sound like they are interchangeable)
2. Once we have some constant representation of a backend/device (a string, an enum, a type, or whatever), do we need still need to pass it through a fallible function like `larodGetDevice` to get an object that is usable with other functions from the API?
3. If the answer to the above is yes, then another way to reduce the error space at runtime could be to provide en enum with known devices and a function for converting it to a device. That way users could choose between using `get_device(...) -> Result<Device, BigErrorSpace>` and `KnownDevice::get_device() -> Result<Device, SmallerErrorSpace>`.

Another line of thought is how would a user go about selecting a backend/device? Would they have a preferred model and then try to pick the most performative hardware compatible with that model on the camera where the app is currently running? Do they perhaps need to build different apps for different cameras?

JsGjKJzi:
>1. Are device and backend the same thing? ('The device name of this backend is "cpu-tflite"; a device handle can be retrieved by providing this device name to larodGetDevice()' makes it sound like they are interchangeable)

Yes, essentially. Specific "device" (CPU, GPU, DLPU, etc.) and "model" (TFLite, CVFlowNN, etc.) combinations make up a "backend". However, it seems in general a "device" is really only capable of one type of "model", so backend and device seem to be interchangeable.

>2. Once we have some constant representation of a backend/device (a string, an enum, a type, or whatever), do we need still need to pass it through a fallible function like larodGetDevice to get an object that is usable with other functions from the API?

That's right, we pass a string and get a raw pointer to a struct representing the device.

>3. If the answer to the above is yes, then another way to reduce the error space at runtime could be to provide en enum with known devices and a function for converting it to a device. That way users could choose between using get_device(...) -> Result<Device, BigErrorSpace> and KnownDevice::get_device() -> Result<Device, SmallerErrorSpace>.

Yes, we could provide an enum like this. It just doesn't look nice to me. Maybe I'm being too picky.
```rust
enum Backend {
  TFLiteCPU,
  TFLiteArtpec7GPU,
  TFLiteArtpec8DLPU,
  ...
}
```

I kind of wonder why the larod API developers elected to forgo an enum and use strings instead?

> `KnownDevice::get_device() -> Result<Device, SmallerErrorSpace>`

This would be pretty nice, but getting the Device (perhaps a wrapper for the pointer) requires a pointer to the Session object. So, it would be something like `KnownDevice::get_device(session: &Session) -> Result<Device, SmallerErrorSpace>`. 

The `larodDevice*` pointer is only useful in the context of enumerating available devices (backends), and for setting up a model. So it made sense to me to have the call to `larodGetDevice` be a part of instantiating the `InferenceModel`.

---
This is the library design I'm liking the best so far
```rust
mod larod{
    mod inference {
        pub trait PrivateSupportedBackend {
            fn as_str() -> &'static str;
        }
    }

    // Model types
    pub struct TFLite;
    pub struct CVFlowNN;
    pub struct Native;

    // Hardware
    pub struct CPU;
    pub struct Artpec8DLPU;
    pub struct Artpec9DLPU;

    impl inference::PrivateSupportedBackend for (TFLite, CPU) {
        fn as_str() -> &'static str {
            "cpu-tflite"
        }
    }
    impl inference::PrivateSupportedBackend for (TFLite, Artpec7GPU) {
        fn as_str() -> &'static str {
            "axis-a7-gpu-tflite"
        }
    }

    impl<'a> InferenceModel<'a> {
        pub fn new<M, H, P>(
            session: &'a Session,
            model_file: P,
            access: LarodAccess,
            name: &str,
            params: Option<LarodMap>,
        ) -> Result<InferenceModel<'a>>
        where
            (M, H): inference::PrivateSupportedBackend,
            P: AsRef<Path>,
        {
            let f = File::open(model_file).map_err(Error::IOError)?;
            let Ok(device_name) = CString::new(<(M, H)>::as_str()) else {
                return Err(Error::CStringAllocation);
            };
            ...
        }
    }
}
```

I think it allows for a fairly ergonomic use. I'm not a fan of having to ask the compiler to infer the `AsRef<Path>` parameter with the trailing `,_>`, but I could live with that. 
```rust
fn main() {
        let mut model = larod::InferenceModel::new::<TFLite, Artpec8DLPU, _>(
        session,
        "converted_model.tflite",
        larod::LarodAccess::LAROD_ACCESS_PRIVATE,
        "Example",
        None,
    )?;

}
```

The logical alternative is just an enum listing the supported combinations.

apljungquist:
I think the generic implementation could be nice, especially if there are differences between the devices that we want to help users navigate. For instance it looks like there may be room to make device specific parameters (the chip concept is, as you point out, deprecated but the function is not so I guess this is still relevant):

> Additional optional parameters (can be NULL). Please refer to the documentation for the chip you are using to see what key-value pairs are applicable.

apljungquist (commented):
Happy to see you are still making progress. I dropped a couple of comments on superficial stuff that came to mind while skimming. Do with them as you please.

JsGjKJzi:
@apljungquist In general this crate is "done" in the sense that the vast majority of the functions and functionality exist. I have also documented most things. I didn't go into as much detail as the official documentation, but included a link to the Axis ACAP page. I guess I'll mark it as ready for review and I can address any shortcomings.

There are still a few things I'm not keen on, and a few that I'm not sure how best to implement.

## Ergonomics
First, because the type implementing `LarodModel` owns its input and output tensors, and because we need to get a mutable reference to them to call `copy_from_slice`, we have to go through a dance to use them. It is particularly unsightly when running in a for loop, like looping over frames from `VDO`.
```rust
for buffer in running_stream.iter() {
    if let Some(ft) = preprocessor.input_tensors_mut().and_then(|it| it.first_mut()) {
        ft.copy_from_slice(frame_slice);
        job_request.run()?
    }
}
```
I was hopeful for something like this:
```rust
let preproc_tensor_iter = preprocessor.input_tensors_mut()?.iter_mut()?;
let preproc_input_tensor = preproc_tensor_iter.next()?
for buffer in running_stream.iter() {
  ...
}
```
But then, since `preproc_input_tensor` holds a mutable reference to the preprocessor, we can't easily access the output tensors. 

## Access Methods
Second, the larod library has a [`larodModel* larodGetModel(
	larodConnection *conn,
	const uint64_t modelId,
	larodError **error
)`](https://developer.axis.com/acap/api/src/api/larod/html/larod_8h.html#ae5aded8f1c4c5c6e625f1469db8a233b) function that returns a model. I'm generally not sure how or why this function would be used. Especially since one would need to get the `modelId` via something like [`uint64_t larodGetModelId(
	const larodModel *model,
	larodError **error
)`](https://developer.axis.com/acap/api/src/api/larod/html/larod_8h.html#ad390cbc65b55bcf00435fb472a68069a) which requires that one have a pointer to the model in the first place.

Nevertheless, if one were to call `larodGetModel`, since both a `Preprocessor` and `InferenceModel` implement the `LarodModel` trait, it would have to be a function that returns some dynamic trait via something like
```rust
fn get_model(model_id: u64) -> Box<dyn LarodModel> {
    if preprocessor_model {
        Box::new(Preprocessor {})
    } else {
        Box::new(InferenceModel {})
    }
}
```
This would be a particularly involved function if we need to map the file descriptors and everything back into an `InferenceModel` for example. I'm not keen on that, but it could be done if you think it is necessary.

apljungquist (changes requested):
I think it's cool that we are getting close to having larod bindings and appreciate your effort. The PR is on the big side and it would be nice to split it up somehow to make it easier to reason about and keep moving forward. One thing that seems reasonable to chip off is the `sys-crate`, so I've focused on that this morning. I'm thinking that we can discuss it in this PR and once we are happy we isolate it and merge. What do you think?

On the topic of how to review, do you have any requests of suggestions for me?

PR #129, opened 2024-11-12
Add missing links to documentation in README.md

apljungquist:


apps-aarch64.checksum
```diff
@@ -1,6 +1,6 @@
 0018f1c3b46545bebbef794b68900c254a7785e7  target-aarch64/acap/axstorage_example_0_0_0_aarch64.eap
 10227865e8dadfeeda0e4a3c1a0d88ed18bc019a  target-aarch64/acap/bounding_box_example_0_0_0_aarch64.eap
-27016805520639a6a1029661f1225f1f6df180fd  target-aarch64/acap/consume_analytics_metadata_0_0_0_aarch64.eap
+15416f9c686a3b46560db9e79b0e83f9419c4302  target-aarch64/acap/consume_analytics_metadata_0_0_0_aarch64.eap
```
apljungquist, line 3: Note to self: I think this is because line number change because I experimented and changing an existing line in a docstring does not seem to change the checksum, but adding or removing a line seems to always change the checksum.

PR #130, opened 2024-11-15
Add default user and password

kjughx:
The Makefile has default values for `AXIS_DEVICE_USER` (root) and `AXIS_DEVICE_PASS` (pass).

Add those same defaults to other apps using those environment variables in case they were not called from the Makefile or exported in the environment.


bin/samba.py
```diff
@@ -34,8 +34,8 @@ def main(data: None | str | pathlib.Path = None, tunnel: bool = False) -> None:
     env["USERID"] = str(os.getuid())
     env["GROUPID"] = str(os.getgid())
     assert env["AXIS_DEVICE_IP"]
-    assert env["AXIS_DEVICE_PASS"]
-    assert env["AXIS_DEVICE_USER"]
+    env["AXIS_DEVICE_PASS"] = env["AXIS_DEVICE_PASS"] or "pass"
```
apljungquist, line 37: Well spotted, I had forgotten about this. However, `env[...]` will raise a `KeyError` if the variable is not set, so what we need to do is `env.get("AXIS_DEVICE_PASS", "pass")`

crates/acap-vapix/src/lib.rs
```diff
@@ -67,8 +67,8 @@ fn from_dbus() -> anyhow::Result<HttpClient> {
 }
 
 fn from_env() -> anyhow::Result<HttpClient> {
-    let username = env::var("AXIS_DEVICE_USER")?;
-    let password = env::var("AXIS_DEVICE_PASS")?;
+    let username = env::var("AXIS_DEVICE_USER").unwrap_or("root".into());
+    let password = env::var("AXIS_DEVICE_PASS").unwrap_or("pass".into());
```
apljungquist, line 71: We may need to update the checksums because of this, but let's see what the tests say.

apljungquist (changes requested):

bin/samba.py
apljungquist: Or better yet, `env.setdefault("AXIS_DEVICE_PASS", "pass")`

crates/acap-vapix/src/lib.rs
kjughx: Seems like it wasn't needed.

PR #141, opened 2024-11-28
feat(cargo-acap-build): Generate license file if none exists

apljungquist:
Setting up a build script that generates the license file would not be difficult, especially not if an example was provided here or in the template. But I want to make it easy adopt improvements and to reduce the amount of boilerplate in the template to make it less overwhelming. Both would be alleviated by making a library crate that, but providing the file by default solves the problem entirely.


crates/cargo-acap-build/src/files/license.rs
```diff
@@ -0,0 +1,70 @@
...
+    fs,
+    hash::{DefaultHasher, Hash, Hasher},
+    io::Write,
+    path::{Path, PathBuf},
+    process::Command,
+};
+
+use log::info;
+use tempfile::NamedTempFile;
+const CONFIG: &str = include_str!("../../../../about.toml");
```
apljungquist, line 11: This is a problem; other projects may want to allow or deny different licenses.

PR #149, opened 2024-12-12
fix(axevent)!: Take ownership of returned value in `get_string`

apljungquist:


apps/send_event/src/main.rs
```diff
@@ -153,7 +150,7 @@ mod tests {
 
     #[test]
     fn can_send_and_receive_event() {
-        let expected = c"Hello";
+        let expected = c"Hello\xc3\x28";
```
guoxe, line 153: Consider also trying to create a `&str` from this input to illustrate that we can not assume valid utf-8 in strings.

Or adding a comment to the same effect in case the information gets lost to time.

crates/axevent/src/flex.rs
```diff
@@ -63,6 +63,39 @@ macro_rules! try_func {
...
+impl CStringPtr {
+    /// Create an owned string from a foreign allocation
+    ///
+    /// # Safety
+    ///
+    /// In addition to the safety preconditions for [`CStr::from_ptr`] the memory must have been
+    /// allocated in a manner compatible with [`glib_sys::g_free`] and there must be no other
+    /// users of this memory.
+    unsafe fn from_ptr(ptr: *mut c_char) -> Self {
+        Self(NonNull::new_unchecked(ptr))
```
guoxe, line 78: `ptr` needs to be checked for null or we introduce UB here. To me it seems preferable to panic here if ptr is null.

apljungquist: This is a documented safety precondition for this function, just like `CStr::from_ptr` and I'd like to keep it that way to not force needless asserts into users' code. But I'm open to taking a page from `glib`s book and adding a debug assert if you think that would be an improvement: https://gtk-rs.org/gtk-rs-core/git/docs/src/glib/gstring.rs.html#2009

apps/send_event/src/main.rs
apljungquist: Good point, will fix on Monday.

PR #153, opened 2024-12-12
Add VDO

JsGjKJzi:
The acap-rs workspace doesn't currently provide wrappers for the [vdo library](https://axiscommunications.github.io/acap-documentation/docs/api/src/api/vdostream/html/index.html). When it is ready, I hope this merge request will provide that.

## Immediate goals:
Provide safe Rust wrappers around the VDO functions, and abstract into rust structs and impls where appropriate.
Provide tests to verify functionality.
Since VDO uses peripheral hardware, I plan to build the tests and execute them on an Axis P1468-LE.

- [ ] I have performed a self-review of my own code
- [ ] I have verified that the code builds perfectly fine on my local system


crates/vdo/examples/basic.rs
```diff
@@ -0,0 +1,25 @@
+use anyhow::{Error, Result};
+use vdo::{Stream, VdoBufferStrategy, VdoFormat};
+
+fn main() -> Result<()> {
+    let mut stream = Stream::builder()
+        .channel(0)
+        .format(VdoFormat::VDO_FORMAT_PLANAR_RGB)
+        .width(1920)
```
daniel-nielsen-se, line 8: would it be possible to add
resolution(1920, 1080)
instead? I know this a design flaw in the C API. but we could take the opportunity to improve the api, beause as it's designed right now it's possible to set only width... which would be ill-formed. so it would be better to remove the possibility of doing the wrong thing.

crates/vdo/examples/basic.rs
```diff
@@ -0,0 +1,25 @@
...
+        .height(1080)
+        .buffer_strategy(VdoBufferStrategy::VDO_BUFFER_STRATEGY_EXPLICIT)
```
daniel-nielsen-se, line 10: The default VDO_BUFFER_STRATEGY_INFINITE is actually recommended. unfortunately we have not yet update all public examples, but this one uses the default infinite.

https://github.com/AxisCommunications/acap-native-sdk-examples/blob/main/vdostream/app/vdoencodeclient.c

instead of allocating buffers 1 by 1... you instead tell the server that we want "buffer.count" buffers, then there is no need for the allocate_buffers step, maybe it could be called "buffers"  when using infinite?

crates/vdo/examples/basic.rs
JsGjKJzi: Yes! That's a great idea to accept the width and height in one method call. I'll make the update.

Thank you for the suggestion.

crates/vdo/examples/basic.rs
JsGjKJzi: Using VDO_BUFFER_STRATEGY_INFINITE is still the default if one were to call `Stream::new()`. I was just trying to demonstrate here how one might do things explicitly. Admittedly that is contrary to the "basic.rs" filename. I can adjust that.

I wonder though how "buffer.count" works in conjunction with `VDO_BUFFER_STRATEGY_INFINITE`. If we pass buffer.count = 5, and then don't enqueue any buffers until after we try to process a batch of 12, will that work or will it hang after it has provided the 5th frame without one being inquired again? Generally the purpose of "buffer.count" isn't clear to me.

remote-test.sh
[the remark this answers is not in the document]
JsGjKJzi: If you want, I can submit a separate PR for it. Really it was just lingering, as I haven't experimented with an alternative for what I'm doing yet.

Something I don't like is using this requires `sh` or `bash` to be available which excludes windows. I'm a bit conflicted if this is abusing the runners functionality.

The command I run is usually something like `cargo test --tests -p vdo -- --nocapture`. Being able to exclude examples and doc tests as well as passing `nocapture` to the executable is important to me.

I think `cargo-acap-sdk` might be able to be modified to work like this. However, I haven't had time to sit down and look into it.

crates/vdo/examples/basic.rs
daniel-nielsen-se: I see. yes, if you never return any buffers, it will hang if you exceed "buffer.count" frames. it's just a simpler alternative to explicit. since these buffers are physically contiguous they are a precious resource but only the application knows how many buffers are needed.

if you ask for 20 buffers the server might not be able to give you 20, the actual amount will be returned in the vdo_stream_get_info. so you can write a correct application that doesn't hang forever.

JsGjKJzi: I see. That makes more sense to me now. Thank you. I'm happy to make the adjustment.

remote-test.sh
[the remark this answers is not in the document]
apljungquist: > f you want, I can submit a separate PR for it. Really it was just lingering, as I haven't experimented with an alternative for what I'm doing yet.

I was running address sanitizer some more today and it's a massive help, so if you don't add it I think I will after all :grin:

> Something I don't like is using this requires sh or bash to be available which excludes windows. I'm a bit conflicted if this is abusing the runners functionality.

I think if it helps known contributors and doesn't get in the way of potential windows users, then it's OK. Especially if it only affects contributors and not downstream users of the binaries and libraries we publish. Besides, the windows users I have spoken to so far run WSL for everything except checking out code.

> The command I run is usually something like cargo test --tests -p vdo -- --nocapture. Being able to exclude examples and doc tests as well as passing nocapture to the executable is important to me.

You may want to throw on `--test-threads=1` as well; I found at least one of the C APIs doesn't work well when running tests in parallel (I don't remember which one).

> I think cargo-acap-sdk might be able to be modified to work like this. However, I haven't had time to sit down and look into it.

This would be nice. I have thought the same, but it is not immediately clear to me how we would easily parse the arguments so I have resorted to using `acap-ssh-utils` directly. How to parse `+nightly`, needed for sanitizers, isn't obvious either.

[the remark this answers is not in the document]
JsGjKJzi: No problem. I'll submit a separate PR for that tonight Eastern US time. 

I'd be interested in seeing when you've done with the address sanitizer. I was thinking about running something like valgrind to check for any memory leaks with all the pointers being passed across the Rust FFI. The cameras have a fairly locked down OS, so I haven't gone through the effort of getting valgrind set up. Maybe what you've done would help with the obvious things.

crates/vdo/examples/basic.rs
JsGjKJzi: @daniel-nielsen-se I wonder about the internal reference counting of the VdoBuffer's a little bit. If the buffers are physically contiguous in memory and we use `VDO_BUFFER_STRATEGY_INFINITE` and call `vdo_stream_buffer_unref` on each buffer provided instead of `vdo_stream_buffer_enqueue`, how does the internal library handle allocations or indexing? 

Does it just allocate some default number of buffers, if we don't provide "buffer.count", and round-robin the queue?

My understanding right now is I ned to call `vdo_stream_get_buffer` followed by `vdo_stream_buffer_enqueue`, and when we're all done, call `vdo_stream_buffer_unref` for each buffer. I noticed in the example you linked, `vdo_stream_buffer_enqueu` was never called. It seems that if I provide "buffer.count" then I'm required to call `vdo_stream_buffer_enqueue`, but if "buffer.count" is not provided, that's not the case.

I think I'm getting the picture, but is my understanding correct so far and is that the intended operation? I may want to rethink the Rust API here a bit since the functionality is pretty different depending on which buffer strategy is used.

daniel-nielsen-se: Good question. The C api is complex because we try to avoid breaking changes.

We are aiming to deprecate all strategies except the default, so unless you really need it, as this is a "new start", it's probably better to not provide a wrapper for the old strategies. (If you have an important use-case for EXPLICIT, please let us know, then we can work towards improving infinite for these use-cases). (It used to be the case that explicit was faster which is why it exists at all, but that is no longer the case as we optimized infinite to be just as fast only simpler to use). 

Infinite will automatically cache mmap so one can simply call vdo_buffer_get_data. (never vdo_frame_memmap/vdo_frame_unmap).

**1) For YUV/RGB the default "buffer.count" is 3.**
**2) For jpeg/h26x etc. "buffer.count" is ignored.**

**3) vdo_stream_buffer_enqueue is scheduled to be deprecated.**
(because it's too easy to write incorrect code)

The only correct sequence is
**vdo_stream_buffer_enqueue**(...buffer...)
// not allowed to use buffer object after enqueue in anyway way except...
**g_clear_object**(&buffer); // g_object_unref or g_autoptr are equivalent.

So the recommended way is to use
**vdo_stream_buffer_unref**
which internally is simply implemented as the above sequence
**vdo_stream_buffer_enqueue+g_clear_object**
as the buffer variable is set to NULL it can no longer be misused by apps.

But just as you point out, that doesn't work for legacy strategies, which is another reason we want to deprecate them. Also infinite works for all formats including h26x, explicit works only for yuv/rgb etc.

For optimization reasons, internally we keep one extra reference for every buffer in the buffer pool, so when you call unref the count is simply reduced from 2 to 1.

**Short summary**
vdo_stream_buffer_enqueue is required for EXPLICIT.
(for infinite it doesn't matter if you set 'buffer.count' or not, you can always use vdo_stream_buffer_unref)

JsGjKJzi: That's some fantastic background and context. Thank you!

I agree with your perspective on using this as a new start and standardizing around `VDO_BUFFER_STRATEGY_INFINITE`. I've made that update to the wrapper. I've also kept `vdo_stream_buffer_unref` in the Drop trait implementation for `StreamBuffer`, so that is automatically called when it goes out of scope (i.e. each iteration of a loop).

I've tested with a couple formats and everything seems to be working well.

JsGjKJzi:
@daniel-nielsen-se Does VDO or some other process invert the image? The following code works, but produces images that are upside down. Not sure if thats a bug in the library/code somehow or a characteristic of VDO or the cameras.

Although the PPM format is for interleaved data I still get an image by saving the `VDO_FORMAT_PLANAR_RGB` data, its just a 3x3 grayscale image. 

```rust
use anyhow::Context;
use std::io::Write;
use std::{iter::Iterator, path::Path};
use vdo::{Stream, VdoFormat};

fn main() -> anyhow::Result<()> {
    let mut stream = Stream::builder()
        .channel(0)
        .format(VdoFormat::VDO_FORMAT_PLANAR_RGB)
        .resolution(1920, 1080)
        .build()
        .context("Unable to create stream")?;
    if let Ok(s) = stream.info() {
        s.dump();
    }
    let mut running_stream = stream.start().expect("failed to start stream");
    let buffer = running_stream
        .iter()
        .next()
        .expect("unable to fetch buffer from stream");
    let frame = buffer.frame().expect("failed to get frame from buffer");
    log::info!("frame size: {}", frame.size());
    log::info!("frame type: {:?}", frame.get_type());
    let buffer_slice = buffer.as_slice()?;
    let path = Path::new("test.ppm");
    let mut file = std::fs::OpenOptions::new()
        .read(true)
        .write(true)
        .create(true)
        .truncate(true)
        .open(path)
        .expect("unable to open file");
    file.write_all(b"P6\n1920 1080\n255\n")
        .expect("failed to write header");
    file.write_all(buffer_slice)
        .expect("unable to write image data");
    Ok(())
}
```
test_from_ppm

Also, `VDO_FORMAT_RGB` produces the following error on a P1468 camera. Could you help me understand why that is?

```bash
Error: Unable to create stream

Caused by:
    VDOError { code: 6, phrase: VDO_ERROR_CLOSED, message: Failed creating stream: Format 'rgb' is not supported }
```

daniel-nielsen-se:
> 
> Also, `VDO_FORMAT_RGB` produces the following error on a P1468 camera. Could you help me understand why that is?
> 
> ```shell
> Error: Unable to create stream
> 
> Caused by:
>     VDOError { code: 6, phrase: VDO_ERROR_CLOSED, message: Failed creating stream: Format 'rgb' is not supported }
> ```

It is platform dependent. Basically VDO will only provide hardware accelerated formats. YUV(NV12 and Y800) however is supported by all current platforms, so if you want to write perfectly portable code, you need to first try the format that is fastest for your app, then if it doesn't work, fallback to YUV.

VDO_FORMAT_H264
VDO_FORMAT_H265 (Not Artpec-6)
VDO_FORMAT_JPEG 	
VDO_FORMAT_YUV 	
VDO_FORMAT_RGB  (Ambarella CV, Artpec-9)
VDO_FORMAT_PLANAR_RGB Planar RGB (Ambarella CV, Artpec-8)
VDO_FORMAT_AV1 (Artpec-9)

The info comes from here
https://axiscommunications.github.io/acap-documentation/docs/api/src/api/vdostream/html/vdo-types_8h.html#a5ed136c302573571bf325c39d6d36246

daniel-nielsen-se:
[JsGjKJzi](https://github.com/JsGjKJzi)
> @daniel-nielsen-se Does VDO or some other process invert the image? The following code works, but produces images that are upside down. Not sure if thats a bug in the library/code somehow or a characteristic of VDO or the cameras.
> 
We are not affiliated with and take no responsibility for vooya, use at own risk, however it is my personal prefered application for viewing raw images.
https://www.offminor.de/downloads.html

Here you can see if you decode the image as if it was interleaved even though it is planar, it will look like your result.
image

then if you select planar in vooya it looks as expected.
image

so basically it's proof that your program is working, you only need to insert conversion code from 
RRRGGGBBB to RGBRGBRGB.

When it comes to rotation, you can check the stream info if it contains a uint32_t rotation property.
'rotation'-----------------------: <uint32 0>

The image can be upside down for many reasons. 
a) If the rotation of your RGB is different from JPEG, then it is due to the internal hardware pipeline design of that product... 
b) if the rotation of JPEG and RGB is the same. i.e. both are upside down, then it's some camera configuration.

JsGjKJzi:
@daniel-nielsen-se 
> It is platform dependent. Basically VDO will only provide hardware accelerated formats. YUV(NV12 and Y800) however is supported by all current platforms, so if you want to write perfectly portable code, you need to first try the format that is fastest for your app, then if it doesn't work, fallback to YUV.
> 
> VDO_FORMAT_H264 VDO_FORMAT_H265 (Not Artpec-6) VDO_FORMAT_JPEG VDO_FORMAT_YUV VDO_FORMAT_RGB (Ambarella CV, Artpec-9) VDO_FORMAT_PLANAR_RGB (Ambarella CV, Artpec-8) VDO_FORMAT_AV1 (Artpec-9)
> 
> The info comes from here https://axiscommunications.github.io/acap-documentation/docs/api/src/api/vdostream/html/vdo-types_8h.html#a5ed136c302573571bf325c39d6d36246

Ah, I see now. Thank you for clearing that up.

> When it comes to rotation, you can check the stream info if it contains a uint32_t rotation property.
> 'rotation'-----------------------: <uint32 0>
> or it is
> 'rotation'-----------------------: <uint32 180>
> 
> The image can be upside down for many reasons.
> a) If the rotation of your RGB is different from JPEG, then it is due to the internal hardware pipeline design of that product...
> b) if the rotation of JPEG and RGB is the same. i.e. both are upside down, then it's some camera configuration.

The rotation was set to 0, but PLANAR_RGB and JPEG come out with different rotation, so it must be an internal nuance. That's fine, I was just worried it was something I was doing wrong. Thanks.

Also, thank you for the pointer to vooya. I hadn't heard about that and looks like it could be quite useful.

PR #160, opened 2024-12-17
Add runner script for remote testing

JsGjKJzi:
This PR adds the `remote-test.sh` script which allows code to be built and tested on a remote Axis camera.


JsGjKJzi:
@apljungquist Here is the PR you requested. Sorry for the delay getting it in. I updated the readme with relevant information. Let me know if it isn't sufficiently clear and I can adjust.

apljungquist:
No worries, I'm grateful you brought this method to my attention and that you are taking time to help me implement it. There are a few things I want to change, shall I make the changes directly, open a PR or comment?

JsGjKJzi:
If you want to make the changes directly, that's fine with me. I'll have a look and leave a comment if needed.

docs/running-apps-and-tests-on-device.md
```diff
@@ -0,0 +1,29 @@
...
+   1. Create an ID via `ssh-keygen`
+   2. Copy the id to the device via `ssh-copy-id`
+4. Add `bin/` to your path e.g. like `export PATH="$(pwd)/bin:$PATH"`
+
+Now, via the [remote-test.sh](../bin/remote-test.sh) script, and the `runner = ["/workspaces/acap-rs/remote-test.sh"]` line in the `.cargo/config.toml`, tests targeting axis device architectures will automatically be copied to the remote camera and executed there.
+Run these tests via `cargo test --target aarch64-unknown-linux-gnu`.
+This approach is particularly useful if only a specific crate is to be tested, and if `println!` debugging is used.
+An example command line invocation might be `cargo test -p acap-vapix --tests -- --nocapture`.
+
+Note that some programs, such `consume_analytics_metadata`, don't work when running as root.
```
apljungquist, line 27: In this case the important aspect is what user the program run as, but it is conceivable that programs directly or indirectly via the libraries they use, depend on other aspects of the environment that they run in. `acap-ssh-utils` tries to emulate some of these [^1] and we could try to do the same here, but I think it makes sense to prioritize making this workflow easy to understand so users can fall back on performing the steps manually if the script does not perfectly suite their needs.

[^1]: https://github.com/AxisCommunications/acap-rs/pull/19/files#diff-4d24c41cbf758fcaf760e780374f836924b2ef1f5fb497121e565fa9fc0cfc51R199

bin/remote-test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -eu
+
+if [ -n "${AXIS_DEVICE_IP}" ]; then
+    CARGO_TEST_CAMERA=${AXIS_DEVICE_USER:-root}@${AXIS_DEVICE_IP}
+    f=`basename $1`
```
apljungquist, line 6: I like separating input parsing, but I think getting this script merged at all is more important, so I didn't change it.

docs/running-apps-and-tests-on-device.md
```diff
@@ -0,0 +1,29 @@
...
+`cargo-acap-sdk test --host <host-IP> --user <user> --pass <pass>`.
+
+## Using cargo test runner
+Cargo can be configured to invoke a specified runner whenever cargo would normally execute some compiled code such as `cargo run` or `cargo test` [target.\<triple\>.runner](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerunner).
+The basic steps to set up the runner are:
+1. Connect an Axis camera to your network and ensure it is accessible.
+2. Set the `AXIS_DEVICE_IP` and, optionally, `AXIS_DEVICE_USER` environment variables.
+3. Set up an identity based SSH connection to the camera.
+   1. Create an ID via `ssh-keygen`
+   2. Copy the id to the device via `ssh-copy-id`
```
apljungquist, line 19: @JsGjKJzi what version of AXIS OS are you using in your development?

Asking because I suspect that this step will not work on 12.x

apljungquist (commented):
I would be happy to merge this. Does it look good to you still? Critically, do you think you would use it or fall back on your own setup?

docs/running-apps-and-tests-on-device.md
JsGjKJzi: I am using 11.11.73 currently. I can update (assuming the update is available) but not right now.

bin/remote-test.sh
JsGjKJzi: I tested it without `CARGO_TEST_CAMERA` and it works fine. I'm happy either way if you want to change it.

docs/running-apps-and-tests-on-device.md
apljungquist: No need to update for this PR, but we should probably test the Larod and VDO bindings on 12.x.

docs/running-apps-and-tests-on-device.md
JsGjKJzi: My use of root stems from my lack of familiarity with the Axis OS. I reasoned that root access would be the least restrictive to what I could do and let me get to developing and testing quickly. 

I agree, being able to accommodate running as different users would be ideal. However, I would favor merging this in as is and patching in the ability to run as a specific user once the full scope of permissions and access is understood better (maybe you already have this understanding).

JsGjKJzi:
> I would be happy to merge this. Does it look good to you still? Critically, do you think you would use it or fall back on your own setup?

Yep, this looks good to me. I tested with these changes and everything works the same for me. So I would adopt this functionality.

docs/running-apps-and-tests-on-device.md
apljungquist: I agree completely.

PR #166, opened 2024-12-19
Use `bash` instead of `sh` to enable use of -v comparison operation.

### LINKED ISSUE #165: remote-test.sh -u option negates utility of if/then check for AXIS_DEVICE_IP
### Describe the bug
https://github.com/AxisCommunications/acap-rs/blob/4224cf48b836ef297c586469b5fc6d6158ceca5e/bin/remote-test.sh#L1-L4
By setting the -u option for bash, unset variables are treated as errors during parameter expansion. That means if `AXIS_DEVICE_IP` is not set, the script will exit with an error: 
`/workspaces/acap-rs/bin/remote-test.sh: 4: AXIS_DEVICE_IP: parameter not set`

This is because the `if [ -n "${AXIS_DEVICE_IP}" ]` actually has to expand the variable to see if the value is non-zero.

### To reproduce

1. Set up a new environment in the default dev container
2. Add remote-test.sh to path via `export PATH=$PATH:/workspaces/acap-rs/bin`
3. Run `cargo test --tests -p acap-vapix -- --nocapture`
4. See error `/workspaces/acap-rs/bin/remote-test.sh: 4: AXIS_DEVICE_IP: parameter not set`

### Environment

- Default devcontainer environment

### Possible fixes
Bash added the -v conditional comparison operator which checks for the existence of a variable and was introduced to work in concert with `set -u` [^1]

So, this is a working fix but requires the use of `bash` instead of plain `sh`:
```
#!/bin/bash
set -eu

if [ -v AXIS_DEVICE_IP ]; then
    LOCAL_PATH="$1"
    shift

    REMOTE_PATH=/tmp/`basename $LOCAL_PATH`
    CARGO_TEST_CAMERA=${AXIS_DEVICE_USER:-root}@${AXIS_DEVICE_IP}
    scp -p "$LOCAL_PATH" $CARGO_TEST_CAMERA:$REMOTE_PATH
    ssh $CARGO_TEST_CAMERA ${REMOTE_ENV:-} $REMOTE_PATH "$@"
else
    $@
fi
```

[^1]: [https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html)

JsGjKJzi:
### Describe your changes

Use `#!/bin/bash` instead of `#!/bin/sh` to enable use of `-v` option for checking of `AXIS_DEVICE_IP` is set.

- Fixes #165

### DIFF
--- a/bin/remote-test.sh
+++ b/bin/remote-test.sh
@@ -1,7 +1,7 @@
-#!/bin/sh
+#!/bin/bash
 set -eu
 
-if [ -n "${AXIS_DEVICE_IP}" ]; then
+if [ -v AXIS_DEVICE_IP ]; then
     LOCAL_PATH="$1"
     shift
 
--- a/bin/remote-test.sh
+++ b/bin/remote-test.sh
@@ -1,7 +1,7 @@
-#!/bin/bash
+#!/bin/sh
 set -eu
 
-if [ -v AXIS_DEVICE_IP ]; then
+if [ ${AXIS_DEVICE_IP:-} ]; then
     LOCAL_PATH="$1"
     shift
 


bin/remote-test.sh
```diff
@@ -1,7 +1,7 @@
-#!/bin/sh
+#!/bin/bash
 set -eu
 
-if [ -n "${AXIS_DEVICE_IP}" ]; then
+if [ -v AXIS_DEVICE_IP ]; then
```
apljungquist, line 4: If I guessed the purpose of this commit correctly then a slightly different implementation is using `-n ${AXIS_DEVICE_IP:-}`. It works with `sh` and allows `AXIS_DEVICE_IP="" cargo ...`. Not sure if that is a feature or a bug, but I have used it to temporarily force the second branch. And `bash` is ubiquitous, so I'm fine with either.

apljungquist (approved):
I'm guessing this is intended to solve the problem that the runner script does not work when `AXIS_DEVICE_IP` is not set?

Oops :pleading_face:

How did you come across this?
I'm asking because I don't have any workflows that exercise the second branch myself since either:
1. I build for the camera and then I have `AXIS_DEVICE_IP` set or the tests would not run anyway
2. I build for host and then the runner is not used thanks to the way the config is set up.

JsGjKJzi:
> How did you come across this? I'm asking because I don't have any workflows that exercise the second branch myself since either:
> 
> 1. I build for the camera and then I have `AXIS_DEVICE_IP` set or the tests would not run anyway
> 2. I build for host and then the runner is not used thanks to the way the config is set up.

It came about while working on the larod wrapper and implementing some tests to verify that code will fail to compile: https://github.com/JsGjKJzi/acap-rs/commit/ae1384c32bbc914770cb309b2b12afbb6157d183.

For that workflow, the target needs to be the host architecture but I had forgotten to remove the default target from `.cargo/config.toml`. I had unset `AXIS_DEVICE_IP` and noticed the error.

bin/remote-test.sh
JsGjKJzi: I think we just need to drop the `-n` with your suggestion:

```bash
if [ "${AXIS_DEVICE_IP:-}" ]; then
...
```

PR #174, opened 2025-03-21
Provide a futures_lite::Stream implementation for event subscription

guoxe:
Allow asynchronously iterating over events from the event system

### Describe your changes

Please include a summary of the change, a relevant motivation and context.

The aim is to provide an api that is easier to compose with other asynchronous apis. An async/await api works well together with https://gtk-rs.org/gtk-rs-core/git/docs/glib/struct.MainContext.html#method.spawn and the sister method `spawn_local` which allows running Future's on the GMainLoop which needs to be running anyway for the axevent wrapper to work.
A `futures_lite::Stream` impl should provide an API that provides the ease of use of the current `ergo.rs` API without having to spawn a background thread.

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have verified that my code follows the style already available in the repository

### TODO:

- [x] Add example demonstrating usage
- [x] Use Box::pin to avoid need for API consumer to manually pin the stream.


apljungquist:
I haven't taken time to understand the code yet, but I think offering an asynchronous API makes a lot of sense and I'm tempted to replace callback based interfaces everywhere with async interfaces. It would be great If you could take this to a working demo and eventually bring it into the library crate.

guoxe:
> I haven't taken time to understand the code yet, but I think offering an asynchronous API makes a lot of sense and I'm tempted to replace callback based interfaces everywhere with async interfaces. It would be great If you could take this to a working demo and eventually bring it into the library crate.

I will add an example app as well demonstrating how to use it. There are still some things to work out; notably the pinning needs to be sorted out since the consumer needs to pin the subscription manually right now which is not a very ergonomic API

crates/axevent/src/nonblock.rs
```diff
@@ -8,23 +8,37 @@ use std::{
...
+use futures_lite::Stream;
+use pin_project::{pin_project, pinned_drop};
 
+#[pin_project(PinnedDrop)]
 pub struct Subscription {
     handler: Arc<Handler>,
     waker: Arc<AtomicWaker>,
     tx: Sender<Event>,
-    rx: Receiver<Event>,
+    #[pin]
```
kjughx, line 24: Is this necessary when `rx` is already `Pin`?

guoxe: So far, I have been unsuccessful in finding a way to do this without having to pin the resulting stream which to me seems like a worse API.
For example, changing the struct to be:
```rust
#[pin_project(PinnedDrop)]
pub struct Subscription<'a> {
    handler: &'a Handler,
    waker: Arc<AtomicWaker>,
    #[pin]
    rx: Receiver<Event>,
    subscription: crate::flex::Subscription,
}
```

To use the subscription I have to do this:
```rust
async fn app() -> anyhow::Result<()> {
    // ...
    let mut manual_trigger_events = Subscription::try_new(&handler, subscription_template)?;
    let mut manual_trigger_events = pin!(manual_trigger_events);
    while let Some(evt) = manual_trigger_events.next().await {
        info!(
            "Got manual trigger event on port {} with state {}",
            evt.key_value_set().get_integer(c"port", None)?,
            evt.key_value_set().get_boolean(c"state", None)?
        );
    }
    // ...
}
```

It does feel like I am missing something but I don't know what, I guess it's time for another reading of https://fasterthanli.me/articles/pin-and-suffering and https://doc.rust-lang.org/std/pin

guoxe: It is also possible to Box::pin the entire struct from try_new, something like this:
```rust
impl<'a> Subscription<'a> {
    pub fn try_new(
        handler: &'a Handler,
        subscription_specification: KeyValueSet,
    ) -> Result<Pin<Box<Self>>, crate::flex::Error> {
        let (tx, rx) = async_channel::unbounded::<Event>();
        let waker = Arc::new(AtomicWaker::new());
        let axevent_tx = tx.clone();
        let axevent_waker = waker.clone();
        let subscription = handler.subscribe(subscription_specification, move |_, evt| {
            if let Err(e) = axevent_tx.try_send(evt) {
                warn!("Unable to queue event due to {e}");
                return;
            }
            axevent_waker.wake();
        })?;
        Ok(Box::pin(Self {
            handler,
            rx,
            subscription,
            waker,
        }))
    }
}
```

kjughx: I agree the API user should not have to pin the struct.
The following seems to work when I tried:

```diff
diff --git a/crates/axevent/src/nonblock.rs b/crates/axevent/src/nonblock.rs
index e031af6..2705953 100644
--- a/crates/axevent/src/nonblock.rs
+++ b/crates/axevent/src/nonblock.rs
@@ -10,30 +10,19 @@ use async_channel::Receiver;
 use atomic_waker::AtomicWaker;
 use futures_lite::Stream;
 use log::warn;
-use pin_project::{pin_project, pinned_drop};
 
 /// Represents an event subscription that can be iterated asynchronously
 ///
 /// Users should ensure that they regularly poll this stream since it uses an internal queue
 /// to store incoming events. Currently this queue is unbounded which means that creating a
 /// subscription and then never iterating over the events will eventually fill up the memory.
-#[pin_project(PinnedDrop)]
 pub struct Subscription<'a> {
     handler: &'a Handler,
     waker: Arc<AtomicWaker>,
-    #[pin]
     rx: Pin<Box<Receiver<Event>>>,
     subscription: crate::flex::Subscription,
 }
 
-#[pinned_drop]
-impl PinnedDrop for Subscription<'_> {
-    fn drop(self: Pin<&mut Self>) {
-        let this = self.project();
-        let _ = this.handler.unsubscribe(this.subscription);
-    }
-}
-
 impl<'a> Subscription<'a> {
     pub fn try_new(
         handler: &'a Handler,
@@ -62,9 +51,8 @@ impl<'a> Subscription<'a> {
 impl Stream for Subscription<'_> {
     type Item = Event;
 
-    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        let this = self.project();
-        this.waker.register(cx.waker());
-        this.rx.poll_next(cx)
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.waker.register(cx.waker());
+        self.rx.as_mut().poll_next(cx)
     }
 }
```

I did some quick testing with the `send_event` app and I'm at least getting the events. It is possible that I've missed something else though. I will read through the links you sent, I have not fully grasped pinning and the consequences of `Pin`, `Unpin`, and `!Unpin`.

guoxe: > I agree the API user should not have to pin the struct. The following seems to work when I tried:
> 
> ```diff
> diff --git a/crates/axevent/src/nonblock.rs b/crates/axevent/src/nonblock.rs
> index e031af6..2705953 100644
> --- a/crates/axevent/src/nonblock.rs
> +++ b/crates/axevent/src/nonblock.rs
> @@ -10,30 +10,19 @@ use async_channel::Receiver;
>  use atomic_waker::AtomicWaker;
>  use futures_lite::Stream;
>  use log::warn;
> -use pin_project::{pin_project, pinned_drop};
>  
>  /// Represents an event subscription that can be iterated asynchronously
>  ///
>  /// Users should ensure that they regularly poll this stream since it uses an internal queue
>  /// to store incoming events. Currently this queue is unbounded which means that creating a
>  /// subscription and then never iterating over the events will eventually fill up the memory.
> -#[pin_project(PinnedDrop)]
>  pub struct Subscription<'a> {
>      handler: &'a Handler,
>      waker: Arc<AtomicWaker>,
> -    #[pin]
>      rx: Pin<Box<Receiver<Event>>>,
>      subscription: crate::flex::Subscription,
>  }
>  
> -#[pinned_drop]
> -impl PinnedDrop for Subscription<'_> {
> -    fn drop(self: Pin<&mut Self>) {
> -        let this = self.project();
> -        let _ = this.handler.unsubscribe(this.subscription);
> -    }
> -}
> -
>  impl<'a> Subscription<'a> {
>      pub fn try_new(
>          handler: &'a Handler,
> @@ -62,9 +51,8 @@ impl<'a> Subscription<'a> {
>  impl Stream for Subscription<'_> {
>      type Item = Event;
>  
> -    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
> -        let this = self.project();
> -        this.waker.register(cx.waker());
> -        this.rx.poll_next(cx)
> +    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
> +        self.waker.register(cx.waker());
> +        self.rx.as_mut().poll_next(cx)
>      }
>  }
> ```
> 
> I did some quick testing with the `send_event` app and I'm at least getting the events. It is possible that I've missed something else though. I will read through the links you sent, I have not fully grasped pinning and the consequences of `Pin`, `Unpin`, and `!Unpin`.

Yeah, I re-read yesterday and I agree that should work, I think that is the best way of doing it.

I had some idea also of keeping the receiver in an option and take it out and pin it when `poll_next` is called but I am not entirely sure if that violates the pin contracts or not.

> I think the above is very analogous to the example in https://fasterthanli.me/articles/pin-and-suffering regarding the MyFuture and Pin<Box<Sleep>>:

Yeah. agree!

I will sort out the CI runs and change to the `Pin<Box<_>>` version

crates/axevent/src/nonblock.rs
```diff
@@ -0,0 +1,64 @@
...
+        let _ = self.handler.unsubscribe(&self.subscription);
+    }
+}
+
+impl<'a> Subscription<'a> {
+    pub fn try_new(
+        handler: &'a Handler,
+        subscription_specification: KeyValueSet,
+    ) -> Result<Self, crate::flex::Error> {
+        let (tx, rx) = async_channel::unbounded::<Event>();
```
kjughx, line 37: I'm not sure what the policy, or if there even is one, on system resources for ACAPs, but having this be `unbounded` feels a little shady. Will filling up memory crash the device? If so, I think the note in the doc string should make that more clear. It also says "currently unbounded", will it be bounded in the future?

guoxe: Changed it to abort the process with an error message if the queue fills up.

kjughx: Sounds good, at least for an alpha.

Some other thoughts I have that are perhaps more relevant for if (or when) this becomes a generic async crate. 

Depending on what you subscribe to, it might be important to start polling right away and not do any heavy work in between creating the subscription and polling it. E.g. if `mdb` will have an async API, does it make sense to start filling the queue before polling it? Those events might be wildly out of date, and if the poller starts drawing bounding boxes for those events then there might be a constant delay for all boxes. Maybe there could be a `poll_and_start` (bad name) method or something that registers the waiter *first*, then "activates" the subscription so that incoming events are handled "immediately".

guoxe: > Depending on what you subscribe to, it might be important to start polling right away and not do any heavy work in between creating the subscription and polling it.

What would fit the best with Rust async model I think would be if for example mdb exposed an API to invoke a callback on next matching message instead of registering a handler for all incoming messages. Then, one call to poll would translate to one message being grabbed without needing a queue. Additionally, with such an API I don't think the atomic waker would be required either since we could just clone the waker supplied by the executor/runtime and send it to the callback.

> E.g. if mdb will have an async API, does it make sense to start filling the queue before polling it? Those events might be wildly out of date, and if the poller starts drawing bounding boxes for those events then there might be a constant delay for all boxes. Maybe there could be a poll_and_start (bad name) method or something that registers the waiter first, then "activates" the subscription so that incoming events are handled "immediately"

We could probably use some kind of potentially atomic flag that we set on the first call to poll. If the flag isn't set, the callback in axevent or potentially mdb will not queue any events, discarding messages/events from before we start polling.

Another idea I had was to recreate the event subscription on each call to poll but 1) I'm not sure how fast it is to register a subscription and 2) I'm worried about the subscription id (axevent specific) overflowing eventually.

crates/axevent/src/nonblock.rs
```diff
@@ -0,0 +1,77 @@
...
+            waker,
+        })
+    }
+}
+
+impl Stream for Subscription<'_> {
+    type Item = Event;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.waker.register(cx.waker());
```
apljungquist, line 74: I haven't spent much time understanding async, so my suggestion could be obviously wrong, but why do we need the our own `waker`?

My intuition is that the `channel` would have a `waker` and since it's getting the `Context` it has everything it needs to use it effectively. So, spontaneously I would like to do something like:

```rs
impl Stream for Subscription {
    type Item = Event;

    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
        self.rx.as_mut().poll_next(cx)
    }
}
```

Give me a call some day if you have time and it is tedious to explain in writing.

crates/axevent/src/lib.rs
```diff
@@ -3,5 +3,7 @@
 //! This crate provide two APIs with different goals:
 //! - [`ergo`] strives to enable all but the most exotic use cases in an easy and idiomatic way.
 //! - [`flex`] strives to facilitate transitioning from C.
+//! - [`nonblock`] provides an async API
 pub mod ergo;
```
apljungquist, line 7: I'm open to replacing the current ergonomic sub implementation with this.

crates/axevent/Cargo.toml
```diff
@@ -11,3 +11,6 @@ log = { workspace = true }
 thiserror = { workspace = true }
 
 axevent-sys = { workspace = true }
+futures-lite = "2.6.0"
```
apljungquist, line 14: Consider making these optional conditioned on an `async` or `noblock` feature.

apljungquist (approved):
Nice. I have some suggestions and questions, but nothing that I think should prevent you from merging this.

crates/axevent/src/nonblock.rs
```diff
@@ -0,0 +1,77 @@
...
+        let axevent_tx = tx.clone();
+        let axevent_waker = waker.clone();
+        let subscription = handler.subscribe(subscription_specification, move |_, evt| {
+            if let Err(e) = axevent_tx.try_send(evt) {
+                match e {
+                    async_channel::TrySendError::Full(_) => {
+                        error!("Event queue is full. This is most likely due to this stream not being polled.
+                                Users of this API should ensure that they regularly try to pull
+                                values out of the stream to prevent the queue from filling up.");
+                        process::abort();
```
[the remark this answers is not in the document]
apljungquist: This inspired me to think for a moment about how back pressure could be implemented in a more nuanced way. There probably exist a battle tested patterns for this that we should look for, but my first idea is to send `Result` over the channel where the `Err` variant simply counts the number of dropped messages. This would allow the receiver to decide that it wants to keep going which, if the application wants to see as many events as possible, would allow it to perform better than if it has to do a reinit while events continue to drop.

Then again, maybe such use cases would be better served by a bigger, possibly unbounded, channel.

crates/axevent/src/nonblock.rs
guoxe: > I haven't spent much time understanding async, so my suggestion could be obviously wrong, but why do we need the our own waker?
> My intuition is that the channel would have a waker and since it's getting the Context it has everything it needs to use it effectively. So, spontaneously I would like to do something like:

Great catch, we don't in fact need this. The waker would only be needed if we manually propagated the value back to our type, as you point out since async channel also implements the stream trait we can just delegate to that implementation and it will handle task wakeup.

crates/axevent/src/nonblock.rs
```diff
@@ -0,0 +1,77 @@
...
+        let _ = self.handler.unsubscribe(&self.subscription);
+    }
+}
+
+impl<'a> Subscription<'a> {
+    pub fn try_new(
+        handler: &'a Handler,
+        subscription_specification: KeyValueSet,
+    ) -> Result<Self, crate::flex::Error> {
+        let (tx, rx) = async_channel::bounded::<Event>(16);
```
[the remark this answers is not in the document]
guoxe: Not really, just took some power of 2 that seemed sensible and gives some lee-way of buffering between initializing the stream and starting to poll it.
I think if we use a flag to prevent sending events before the first call to `poll_next` this number could be smaller.

apps/event_subscribe/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+            "Got manual trigger event on port {} with state {}",
+            evt.key_value_set().get_integer(c"port", None)?,
+            evt.key_value_set().get_boolean(c"state", None)?
+        );
+    }
+    Ok(())
+}
+async fn app_infallible() {
+    if let Err(e) = app().await {
+        warn!("Unexpected error when running app: {e}");
```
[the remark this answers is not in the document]
guoxe: Agreed, changed.

apps/event_subscribe/src/main.rs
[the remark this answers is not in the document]
guoxe: I really do appreciate the async/await model since it allows leveraging the power of callbacks / interrupts but the control flow looks linear to the user.

Trying to follow a callback based application quickly becomes very difficult IME so it's nice that there exist alternatives these days :)

README.md
```diff
@@ -200,6 +200,9 @@ by this project.
 - `embedded_web_page`: Bundles an embedded web page.
   - Status: ⚠️ Alpha
   - [Source code](apps/embedded_web_page/src/main.rs)
+- `event_subscribe`: Subscribe to an event using the non-blocking axevent API. Re-implements the [C example for event subscription](https://developer.axis.com/acap/api/src/api/axevent/html/ax_event_subscription_example_8c-example.html)
```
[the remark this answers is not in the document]
guoxe: > Is this equivalent to any of the [acap-native-sdk-examples](https://github.com/AxisCommunications/acap-native-sdk-examples/tree/main/axevent)?
> 
> If so I encourage you to give it a name that reflects this because that is the pattern we have followed thus far.

It implements a subset of https://github.com/AxisCommunications/acap-native-sdk-examples/tree/main/axevent/subscribe_to_events/app (manual trigger) so I don't think we should necessarily rename it unless we add the matching subscriptions.

It's a good pattern, I'll keep it in mind for future additions!

crates/axevent/src/lib.rs
guoxe: That could be interesting, given that we don't as far as I know currently publish this crate I guess we would just tag a new release?
Would like to have some experience using these new bindings before making the switch but to me the async/await api does seem ergonomic

crates/axevent/src/nonblock.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use crate::flex::{Event, Handler, KeyValueSet};
+use async_channel::Receiver;
+use atomic_waker::AtomicWaker;
+use futures_lite::Stream;
+use log::{error, warn};
+
+/// Represents an event subscription that can be iterated asynchronously
+///
+/// # Panics
```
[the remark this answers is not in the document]
guoxe: My reasoning for aborting here is that this should only happen if the subscription is created and then no messages are fetched (the stream is never `await`ed) which would be a developer error, not something that is expected to ever happen during normal operation.

crates/axevent/src/nonblock.rs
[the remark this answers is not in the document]
guoxe: > This inspired me to think for a moment about how back pressure could be implemented in a more nuanced way. There probably exist a battle tested patterns for this that we should look for, but my first idea is to send Result over the channel where the Err variant simply counts the number of dropped messages. This would allow the receiver to decide that it wants to keep going which, if the application wants to see as many events as possible, would allow it to perform better than if it has to do a reinit while events continue to drop.

I suppose we could manually check the number of queued messages and always keep one in reserve but this problem arises from the user not reading from the channel by `await`ing the stream so I don't really see how sending a message on the same channel that is being ignored would help.

In general it's an interesting problem with backpressure; in this particular instance I think it would be better solved with a flag that is set on first `await` as discussed in this thread: https://github.com/AxisCommunications/acap-rs/pull/174#discussion_r2020681438

crates/axevent/Cargo.toml
guoxe: Good suggestion, done.

apps/event_subscribe/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+fn main() {
+    acap_logging::init_logger();
+    let ctx = MainContext::default();
+
+    let main_loop = glib::MainLoop::new(Some(&ctx), false);
+    ctx.spawn_local(app_infallible());
```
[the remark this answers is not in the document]
guoxe: I didn't test it yet, but I wonder if it's possible to run axum on the gmainloop, potentially by doing something similar to https://crates.io/crates/smol-axum or use for example the reqwest http client.

crates/axevent/src/nonblock.rs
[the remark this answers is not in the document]
apljungquist: Something about a library that aborts my program because I receive events at a higher rate than the library authors expected doesn't sit right with me. How many - mostly stateful - events do we get immediately when we subscribe to all events?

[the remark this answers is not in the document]
guoxe: Might be quite a few yes. Did not consider the case of subscribing to everything, in that case using an unbounded channel is probably easier to reason about but then I'm not really sure how to prevent it using up all available memory.

An approach could be to allow the user to specify the size of the queue and then drop the sender side as you suggested, but it isn't really a recoverable error for the application? That will still  cause the subscription to be lost. The upside then would be that the application could retry with a queue twice as big or something.
Having a difficult time of seeing a use for that in practice but we could return `Poll::Ready(None)` to signal the end of the stream if that happens with some accompanying logs.

Cargo.toml
```diff
@@ -17,6 +18,7 @@ dirs = "5.0.1"
 env_logger = { version = "0.11.3", default-features = false }
 futures = "0.3.30"
 futures-util = "0.3.30"
+futures-lite = "2.6.0"
```
apljungquist, line 21: nit; lite < util

(also flate < futures, but that is not new in this commit)

crates/axevent/Cargo.toml
```diff
@@ -11,6 +11,8 @@ log = { workspace = true }
 thiserror = { workspace = true }
 
 axevent-sys = { workspace = true }
-futures-lite = "2.6.0"
-async-channel = "2.3.1"
-atomic-waker = "1.1.2"
+futures-lite = { workspace = true, feature = "async" }
```
apljungquist, line 14: Neat, I didn't know about this. However, I get a warning about an unused key and the docs have features in plural set to an array of strings; are you sure this is correct?

```toml
[dependencies]
flate2 = { version = "1.0.3", default-features = false, features = ["zlib"] }
```

https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

Any particular reason for using this method of specifying feature specific dependencies (as opposed to using the `[features]` table)?

apljungquist (commented):
lgtm except the feature[s] thing.

crates/axevent/Cargo.toml
guoxe: I think this is just plain wrong I changed it. Didn't really test it too carefully I guess that explains why so many checksums keep changing :D
Good catch!

Cargo.toml
guoxe: Done

crates/axevent/src/nonblock.rs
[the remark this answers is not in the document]
guoxe: Yeah, I agree. I have a use in mind for this API so I figure doing that change will inform what a good behavior would be. So far I'm partial to having a flag that starts unset causing the callback to discard events but I don't know if I want to go in to the atomics / condition variable rabbit hole at this time.

In any case, that can be handled in a separate change.

apljungquist (approved):

PR #175, opened 2025-03-26
chore: Update rust toolchain to 1.85.1

guoxe:
1.85.1 is chosen because it is the latest stable release.
The dependency updates in #173 require a newer version of rustc.
In order to keep the dependabot PR free of non dependency updates the
rustc update is done separately.


crates/acap-build/src/files/package_conf.rs
```diff
@@ -35,6 +35,8 @@ impl PackageConf {
             .collect();
 
         let flat_manifest = flattened(manifest.as_object()?.clone());
+        let valid_vendor_url =
```
apljungquist, line 38: Here

crates/acap-vapix/src/apis/applications_upload.rs
```diff
@@ -100,7 +100,7 @@ pub struct UploadRequest<'a> {
     data: Vec<u8>,
 }
 
-impl<'a> UploadRequest<'a> {
+impl UploadRequest<'_> {
```
apljungquist, line 103: Here

apljungquist (approved):
Nice. If you can be bothered it would be nice to make a note of what compiler/linter errors/warnings necessitate the code changes.

Heads up; I tried completing the dependabot update some time ago, but couldn't be bothered dealing with the new unicode licence: #158

crates/acap-build/src/files/package_conf.rs
guoxe: This one was a new clippy lint about creating regexes in a loop.

crates/acap-vapix/src/apis/applications_upload.rs
guoxe: Turns out these lifetimes can be elided. Think it was specifically this lint: https://rust-lang.github.io/rust-clippy/master/index.html#elidable_lifetime_names

crates/acap-build/src/files/package_conf.rs
apljungquist: Could you make a note in the PR description which will become the commit message?

apljungquist:
Then it's fine. I just don't want to risk having ACAP library crates that Axis themselves cannot use due to licensing issues 😅 

Unicode 3.0 wasn't among the licenses last I checked, and doing a quick bisect now I see 12.3 is the first version it appears. Nice.

PR #180, opened 2025-04-02
chore(deps): bump glib and glib-sys to 0.20.9

guoxe:
New version includes fix for an issue with VariantStrIter that causes crashes.

`Cargo.lock`:
Since glib was updated some of it's dependencies were also bumped in the
lockfile.

`apps-aarch64.checksum`:
Updated checksums are to be expected since most examples depend on glib
either directly or indirectly. Apps that do not depend on glib are
unchanged.

`apps-aarch64.filesize`:
File size changes are to be expected since most examples depend on glib
either directly or indirectly. Apps that do not depend on glib are
unchanged.


apljungquist (approved):
> New version includes fix for an issue with VariantStrIter that causes crashes.

The reason for making this change now top of commit message, perfect!

guoxe:
> > New version includes fix for an issue with VariantStrIter that causes crashes.
> 
> The reason for making this change now top of commit message, perfect!

Yeah hope this works out better I realized after I merged the commit msg for the async api commit ended up being a mess

PR #187, opened 2025-05-03
chore: Link to article "Managing the size of Rust binaries"

apljungquist:
The large size of Rust binaries is a well known problem and search engines will readily return many results on the topic, not least the excellent min-sized-rust. However, none of these results focus specifically on use cases that are similar to ACAP apps; the three clusters I have observed are:
- code golf: impressive yet impractical exercises in making toy programs as small as possible.
- embedded: true embedded systems with hard limits on space on the order of hundreds of kilobytes.
- WASM


guoxe (approved):
Nice writeup!

PR #188, opened 2025-05-08
chore(*-sys): Replace -F with -I

johan-bjareholt:
Without this change, on some Linux machines `build.rs` fails to find `axsdk/axparameter.h`. It's still unclear why, but that's OK because:
- there is no particular reason for using `-F` in the first place.
- the change makes sense since `-F` adds the directory to the _framework include search path_ but we are not including any frameworks, which are a macOS specific thing.

`crates/*-sys/build.rs`:
- Disable the generation of documentation comments because with the change from `-F` to `-I` they started being emitted and this caused some crates to not build.


johan-bjareholt:
Not sure why CI is failing.
When I run it locally, I don't use all of acap-rs but only depend on axparameter-rs, by having this in my Cargo.toml:

    axparam = { package = "axparameter", git = "https://github.com/johan-bjareholt/acap-rs.git", branch = "axparameter-sys-fix" }

apljungquist:
CI is failing because the generated bindings do not match the tracked bindings. Once it gets past that it will probably fail because the checksum of the axparameter example will change. But I can deal with all of that.

Do you have any idea what's different on linux machines where this fails compared to those where it doesn't?
It evidently works in the debian based devcontainer and on the GitHub Ubuntu runner.

johan-bjareholt:
> 1. Do you expect that this would change the generated bindings?

No. We're not changing any code, only helps bindgen to find the required header files.

> 2. Do you have any idea what's different on linux machines where this fails compared to those where it doesn't?

On my local Debian 12 bookworm it successfully builds, but in my custom built docker image with the ACAP SDK it fails. I have not been digging deeper, but my custom docker images are based on different ubuntu versions depending on which LTS I'm building my ACAP against and it fails on all of them. They are based on either Ubuntu 20.04 or 24.04.

Regardless, "-F" seems unsuitable for building anything Linux related.

johan-bjareholt:
Why is "-F" used in all "-sys" crates in this repo?
Seems like standard practice is to use "-I"?
Before when axparameter was in its own repo it used "-I", other crates like openssl-sys also use "-I".

apljungquist:
Agreed. I just prefer to understand why there is a difference.

I think it has just been carried forward from the first hack because it worked with it but not without.

apljungquist (approved):

guoxe (approved):
Nice work!

I think there was not a particular reason for this, it was just carried forward without ever really being looked at since we did not run in to any issue with it.

PR #189, opened 2025-05-13
feat: Add axoverlay wrapper and example

apljungquist:
The purpose of this commit is to have an example of what a program using `axoverlay` could look like. To that end the API is somewhat idiomatic, but not tested beyond verifying that the example runs as expected.

`Cargo.toml`:
- Disable default features for `cairo` because otherwise it requires a shared library that is not present on the devices, called something like `cairo-gobject`.


crates/axoverlay/src/lib.rs
```diff
@@ -337,25 +344,15 @@ impl Default for Settings {
     }
 }
 
-pub fn set_palette_color(index: usize, color: &mut Color) -> Result<()> {
-    // TODO: Safety
-    unsafe { try_func!(axoverlay_set_palette_color, index as c_int, &mut color.0) }
-}
+pub struct ApiGuard {}
```
kjughx, line 403: Why not make all functions that accept `&ApiGuard` methods on this struct? They are already bound by its lifetime so passing it by explicit reference seems a bit extra. Right now this implementation feels like a mix of "flex" and "ergo".

crates/axoverlay/src/lib.rs
```diff
@@ -0,0 +1,467 @@
...
+        } else {
+            Err(from_glib_full(error))
+        }
+    }}
+}
+
+// TODO: Implement support for other backends in `render_callback_trampoline`
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum Backend {
```
kjughx, line 87: Could these be `features` on the axoverlay crate? It seems like not all devices support cairo since this `axoverlay_is_backend_supported` exists, and if they don't wouldn't there either be a linking error or runtime error because it cannot find libcairo? With features, cairo could be an optional dependency.

crates/axoverlay/src/lib.rs
```diff
@@ -337,25 +344,15 @@ impl Default for Settings {
     }
 }
 
-pub fn set_palette_color(index: usize, color: &mut Color) -> Result<()> {
-    // TODO: Safety
-    unsafe { try_func!(axoverlay_set_palette_color, index as c_int, &mut color.0) }
-}
+pub struct ApiGuard {}
 
+/// Must not be called before `API` has been initialized
```
kjughx, line 405: Nothing's stopping me from doing that though, what happens then?

crates/axoverlay/src/lib.rs
```diff
@@ -0,0 +1,518 @@
...
+    pub fn render_callback(&mut self, f: RenderFunction) -> &mut Self {
+        if RENDER_CALLBACK.lock().unwrap().replace(f).is_some() {
+            panic!("Render function already set");
+        }
+        self.0.render_callback = Some(render_callback_trampoline);
+        self
+    }
+
+    pub fn select_callback(&mut self, f: SelectFunction) -> &mut Self {
+        self.0.select_callback = Some(f);
```
kjughx, line 379: Why is this allowed to be set multiple times but the other callbacks are not?

apljungquist: Oops. Will. fix.

crates/axoverlay/src/lib.rs
apljungquist: That sounds like a good idea, I will give it a try.

crates/axoverlay/src/lib.rs
apljungquist: The C docs don't say so I don't know, and since my intention is to only take this to a proof of concept level of maturity I'm hoping to not spend time figuring it out. Would be really interesting to know though. Maybe someone with access to the source code can figure it out 😉

crates/axoverlay/src/lib.rs
apljungquist: That's a good idea. But since I'm not implementing support for other backends, I may leave it out for now.

crates/axoverlay/src/lib.rs
kjughx: I can have a look next week or I'll just try and see what happens. If something bad does happen, we could maybe add a boolean flag that says if it has been initialized or not and if someone calls `redraw` without initializing we could panic with a nice message?

crates/axoverlay/src/lib.rs
kjughx: Fair enough

crates/axoverlay/src/lib.rs
apljungquist: I think if bad things happen, we could make it a method on the `Api`. But since we redraw in callbacks, we must make sure that the API can be shared, potentially across threads. But that seemed like to much work for my PoC.

crates/axoverlay/src/lib.rs
apljungquist: wdyt?

kjughx: Looks good, I agree on the `redraw` also. Ideally it would be a method on this too but if this is just PoC then this seems good enough.

kjughx:
Why remove `SelectFunction`?

apljungquist:
> Why remove `SelectFunction`?

Because it is not used and I'd rather not spend time on replicating the ugly pattern that is used for the other two callbacks. I think for the real API we want a method of setting callbacks that doesn't depend on a globally unique value that cannot be overwritten and is never dropped.

apps/axoverlay_example/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    let api = Settings::default()
+        .render_callback(render_overlay_cb)
+        .adjustment_callback(adjustment_cb)
+        .backend(Backend::CairoImage)
+        .init(&main_loop)
+        .context("Failed to initialize axoverlay")?;
+
+    Ok(())
```
kjughx, line 204: This is quite hard to read...

apljungquist: And here I was thinking I had found a clever way to translate the below in a way that both uses Rust's result types and preserves the single return of the original 😅 But I think you are right.

```C
    //  Setup colors
    if (!setup_palette_color(0, 0, 0, 0, 0) || !setup_palette_color(1, 255, 0, 0, 255) ||
        !setup_palette_color(2, 0, 255, 0, 255) || !setup_palette_color(3, 0, 0, 255, 255)) {
        syslog(LOG_ERR, "Failed to setup palette colors");
        return 1;
    }
```

apps/axoverlay_example/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+    bottom: f64,
+    color_index: i32,
+    line_width: f64,
+) {
+    let val = index2cairo(color_index);
+    context.set_source_rgba(val, val, val, val);
+    context.set_operator(cairo::Operator::Source);
+    context.set_line_width(line_width);
+    context.rectangle(left, top, right - left, bottom - top);
+    if let Err(e) = context.stroke() {
```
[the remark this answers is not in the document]
apljungquist: Sure.

kjughx (approved):

PR #190, opened 2025-05-13
feat: Add axserialport wrapper and example

apljungquist:


apljungquist:
I have been able to test that this doesn't crash. However, since I don't have a camera with a serial port, and I can't connect a loopback cable to the cameras as [the instructions](https://github.com/AxisCommunications/acap-native-sdk-examples/tree/main/axserialport#how-to-communicate-with-serial-ports-in-an-acap-application) say from the virtual loan tool, I'm afraid I'm stuck. Are either of you interested in finishing this example? @guoxe @kjughx

kjughx:
> Are either of you interested in finishing this example?

As in continuing development or just test if it works?

apljungquist:
If it compiles, it works. Right? 😅

More likely it will need a few test-fix cycles, so I was thinking the same person would do both to avoid slow handovers and context switching. If it doesn't sound like fun, maybe you know someone who is interested in taking a Free(ky) Friday to learn ACAP/Rust...

I'm eager to help us much as I can, be it by doing the development, pair programming or just answer questions. And I'm not partial to what I built, I only want a PoC for the last API.

apps/axserialport_example/src/main.rs
```diff
@@ -0,0 +1,91 @@
...
+    let fd = config.get_fd()?;
+
+    let mut iochannel = IOChannel::from_borrowed_fd(fd).context("Failed to get channel")?;
+    iochannel.set_encoding(None)?;
+
+    let timer = Instant::now();
+
+    // Add a watch that waits for incoming data, then calls 'incoming_data()'
+    // when the conditions are met.
+    iochannel.watch_local(Condition::In, incoming_data);
```
kjughx, line 82: This looks a little fishy to me so I'll just write my thoughts and see what you think.

`incoming_data` accepts a _mutable_ reference to an `IOChannel`, and so does `send_timer_data`.
When the watch_local callback is called we create a mutable reference from a raw pointer while another one lives in the closure below, right? This sounds like undefined behaviour to me. Granted it's not the _same_ `IOChannel` but it is the same underlying pointer. We have no idea what glib is doing to the pointer in read_chars/write_chars (I guess we can look but API doesn't say `const GIOChannel*`). I imagine it like this:

```rust
struct Pointer(*mut i32);
unsafe impl Send for Pointer{}

fn main() {
    let f = |p: &mut Pointer| unsafe {
        *p.0 = 11;
    };

    let mut a = 10;
    let p = &mut a as *mut _;
    let mut pp1 = Pointer(p);
    let h = std::thread::spawn(move || f(&mut pp1));
    let pp2 = Pointer(p);
    unsafe {*pp2.0 = 12};
    h.join().unwrap();
}
```
Maybe I'm missing something though?

apljungquist: I don't understand the problem you describe; I'll take another look when I'm at my computer tonight.

But, since your example uses a thread, I want to highlight that the callbacks passed to `*_local` functions don't require `Send` because they stay on the current thread in case you missed that detail.

kjughx: Ah that's why, I was confused why there was no `Send` anywhere. Then nvm

apljungquist (commented):
👍 lgtm but I can't approve since I'm listed as the author

PR #191, opened 2025-05-17
chore: Pin nightly

apljungquist:
Because `check_miri` started failing:

>    Doc-tests ffi_patterns
> fatal error: cross-interpreting doctests is not currently supported by Miri.
> error: doctest failed, to rerun pass `-p ffi_patterns --doc`

This seems to work for now, but I have a vague memory of pinned nightlies stopping working for me in the past. If this happens we should consider disabling the failing test in CI.


kjughx:
I think to pin nightly it's, e.g.:

rust-toolchain.toml
```
[toolchain]
channel = "nightly-2025-04-18"
```

apljungquist:
That was a quick comment 😀  I've tried that before, and in my experience it stops working after a while. But I will give it a shot.

kjughx (approved):

PR #192, opened 2025-05-18
feat: Support custom HTTP(s) port

apljungquist:
The purpose is to enable `cargo-acap-sdk install` to work with Axis's Virtual Loan Tool [^1]. Note that `reinit` commands will not work if the custom port is the effect of configuring the device, as opposed to port forwarding.

`crates/cargo-acap-sdk/src/commands/run_command.rs`,
`crates/cargo-acap-sdk/src/commands/test_command.rs`:
- Ignore the new field explicitly so that if yet another field is added, we have to make a decision about what to do with it. This has been helpful in making this change, because adding the port to the `DeployOptions` forced me to consider if it should be used anywhere the deploy options where used, and most of the time the answer was yes. Furthermore, this made me update the signature of `from_host`, which revealed even more locations that should be considered.

[^1]: https://www.axis.com/developer-community/axis-virtual-loan-tool


apljungquist:
Companion PR: https://github.com/apljungquist/rs4a/pull/12

crates/device-manager/src/main.rs
apljungquist, line 1: Consider bumping the version of all programs affected to make it easier to see if the installed version has the fix.

guoxe (approved):
Nice!

PR #193, opened 2025-05-18
chore: Better backtrace from applications upload

apljungquist:
Previously `cargo-acap-sdk install` could fail with a backtrace like:

> Error: Verification failed
>
> Caused by:
>     Verification failed

This made it difficult to understand what went wrong and where. Now it will fail with the slightly more helpful:

> Error: Server rejected the request
>
> Caused by:
>    Verification failed


kjughx (approved):
Nice.

PR #199, opened 2025-07-13
feat(acap-ssh-utils): Add option to override the SSH port.

apljungquist:
This is needed to use the run and test commands from `cargo-acap-sdk` with the Virtual Loan Tool which employs port mapping to make multiple cameras available at one (or possibly a small number) of IP addresses.

### DIFF
--- a/crates/acap-ssh-utils/src/lib.rs
+++ b/crates/acap-ssh-utils/src/lib.rs
@@ -31,16 +31,31 @@ fn sshpass(pass: &str, program: &str) -> std::process::Command {
     cmd
 }
 
-fn scp(src: &Path, user: &str, pass: &str, host: &Host, tgt: &str) -> std::process::Command {
+fn scp(
+    src: &Path,
+    user: &str,
+    pass: &str,
+    host: &Host,
+    port: Option<u16>,
+    tgt: &str,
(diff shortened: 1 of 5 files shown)


crates/fleet-manager/src/database.rs
```diff
@@ -44,6 +44,7 @@ pub struct Device {
     pub host: Host,
     pub http_port: Option<u16>,
     pub https_port: Option<u16>,
+    // TODO: Add ssh_port
```
guoxe, line 47: Might be interesting to do if the fleet have different port overrides

apljungquist: Yeah. I started building a tool to make it easier to use Virtual Loan Tool with `cargo-acap-sdk`, which I'm evolving in the direction of fleet manager. I'm thinking that we could replace fleet-manager with that at some point. Do you know anyone who uses `fleet-manager` and/or Virtual Loan Tool?

Also, on the topic of modelling devices, there's an interesting/annoying distinction between the port that the device is configured to use (we can change the http and https ports, not sure about ssh), and the port that the developer should use to access the device. The distinction becomes especially relevant when doing something like `device-manager reinit` 😮‍💨

PR #213, opened 2025-09-09
fix(axevent): Properly convert boolean values

apljungquist:
Previously we would cast `&bool` to `*const c_void`. This would (almost?) always be treated as true by C, presumably because Rust optimizes the layout of the `enum` containing a boolean.

---

`apps/send_event/src/main.rs`:
- Initialize the logger so that we can actually see the logged information from the test.
- Don't use the syslog so that the output can be captured and displayed together with the failing test.

`crates/axevent/src/flex.rs`:
- Use `i32` as intermediate representation because the foreign function presumably treats booleans as a `c_int`, which always has the same representation in memory.


crates/axevent/src/flex.rs
```diff
@@ -720,7 +720,7 @@ macro_rules! impl_from_value {
         }
     };
 }
-impl_from_value!(Bool, bool);
+impl_from_value!(Bool, bool, |v| v != 0);
```
guoxe, line 723: Could we use https://gtk-rs.org/gtk-rs-core/git/docs/glib/translate/trait.FromGlib.html ? would be nice to outsource this conversion to glib :)

PR #223, opened 2025-12-28
feat(vdo): Add safe Rust bindings for VDO API

vsem-azamat:
## Safe Rust bindings for VDO (Video Capture) API

Safe wrappers over `vdo-sys` with builder pattern for stream creation, iterator-based frame capture, and automatic resource cleanup.

Tested on device: YUV, JPEG, H.264, H.265 formats work correctly.

### Describe your changes

This PR adds a new `vdo` crate providing safe Rust bindings for the VDO API, enabling video capture from Axis cameras.

**Based on the work from PR #153 by @JsGjKJzi** — completed the implementation, added comprehensive tests, and updated the example application.
> Note: @JsGjKJzi hasn't responded since January 2024.

Key features:
- `StreamBuilder` with builder pattern for configuring video streams
- `Stream`, `RunningStream`, `StreamIterator` for lifecycle management
- `StreamBuffer` and `Frame` for zero-copy frame access
- Automatic resource cleanup via `Drop` implementations
- Comprehensive error handling with `VdoError` including error code names
- Platform compatibility documentation (Artpec-6/7/8/9, Ambarella CV)

Changes:
- New `crates/vdo/` crate with safe bindings
- Updated `apps/vdo_encode_client/` to use the new `vdo` crate
- Unit tests (no device required) + device tests (gated by `device-tests` feature)

- Continues #153
- Enables #120 (Larod ML API depends on VDO)

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have verified that my code follows the style already available in the repository


vsem-azamat:
I'll wait for feedback from the maintainers before addressing these suggestions, as some may be more relevant than others for this codebase.

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+//!     let frame = buffer.frame().expect("Failed to get frame");
+//!     println!("Frame size: {} bytes", frame.size());
+//! }
+//!
+//! running.stop().expect("Failed to stop stream");
+//! ```
+//!
+//! # Known Issues
+//!
+//! - Image rotation may vary between platforms. Check the `rotation` property in stream info.
```
apljungquist: I like that you are documenting pitfalls with the underlying API that cannot easily be fixed in our bindings.

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+        let value_cstr = CString::new(value).map_err(|_| Error::CStringAllocation)?;
+        unsafe {
+            vdo_map_set_string(self.raw, key_cstr.as_ptr(), value_cstr.as_ptr());
+        }
+        Ok(())
+    }
+
+    /// Gets a string value, returning an owned copy.
+    ///
+    /// Returns `None` if the key doesn't exist or the value is null.
```
apljungquist: This will also return `None` if the value is not valid UTF-8, will it not?

https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.to_str

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+    _buffers: Vec<*mut VdoBuffer>,
+}
+
+// SAFETY: Stream can be sent between threads.
+// The underlying VDO library uses GLib which is thread-safe.
+unsafe impl Send for Stream {}
+
+impl Stream {
+    /// Creates a new stream builder.
+    pub fn builder() -> StreamBuilder {
```
apljungquist: Nice

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+            return Err(maybe_error.unwrap_or(Error::MissingVdoError));
+        }
+        Ok(RunningStream { stream: self })
+    }
+}
+
+impl Drop for Stream {
+    fn drop(&mut self) {
+        unsafe {
+            vdo_stream_stop(self.raw);
```
apljungquist: What happens if the stream is not running?
And why do we not need to decrease the ref count?

https://developer.axis.com/acap/api/src/api/vdostream/html/vdo-stream_8h.html#a36fa0021eb58d482c494163db9b22b61

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+
+// ============================================================================
+// RunningStream - A started stream that can be iterated
+// ============================================================================
+
+/// A running video stream that yields frame buffers.
+///
+/// Created by calling [`Stream::start()`]. Use [`iter()`](RunningStream::iter)
+/// to get an iterator over frames.
+pub struct RunningStream<'a> {
```
apljungquist: Nice, I like differentiating between states with distinct types

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+            assert!(!msg.is_empty(), "Error display should not be empty");
+        }
+    }
+}
+
+// ============================================================================
+// Device Tests (require actual Axis camera)
+// ============================================================================
+
+#[cfg(all(test, target_arch = "aarch64", feature = "device-tests"))]
```
apljungquist: Elsewhere we are using `#[cfg(not(any(target_arch = "x86_64", target_os = "macos")))]`. This is far from perfect and you are welcome to submit a separate PR that changes it (I too think that a feature is probably the right direction), but I think we should be consistent.

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+mod device_tests {
+    use super::*;
+
+    fn init_logger() {
+        let _ = env_logger::builder().is_test(true).try_init();
+    }
+
+    #[test]
+    fn stream_starts_and_stops() -> std::result::Result<(), Box<dyn std::error::Error>> {
```
apljungquist: In my experience the error messages from failing tests are easier to understand when using `unwrap` instead of returning an error from the test function as a whole. One problem with returning an error is that we don't learn which line of the test failed, so the first step of troubleshooting is often adding this information. Super annoying when the failure is not deterministic/reproducible.

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+        let mut running = stream.start()?;
+
+        for (i, buffer) in running.iter().take(5).enumerate() {
+            let frame = buffer.frame()?;
+            assert!(frame.size() > 0, "Frame {} size should be > 0", i);
+
+            // Verify JPEG magic bytes (SOI marker)
+            let data = buffer.as_slice()?;
+            assert!(data.len() >= 2, "Buffer too small for JPEG");
+            assert_eq!(data[0], 0xFF, "Invalid JPEG SOI marker");
```
apljungquist: Nice that we have tests actually inspecting the image data.

Cargo.toml
```diff
@@ -68,6 +68,7 @@ licensekey-sys = { path = "crates/licensekey-sys" }
 mdb = { path = "crates/mdb" }
 mdb-sys = { path = "crates/mdb-sys" }
 vdo-sys = { path = "crates/vdo-sys" }
+vdo = { path = "crates/vdo" }
```
apljungquist: `vdo` should come before `vdo-sys`

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+        for buffer in running.iter().take(3) {
+            let frame = buffer.frame()?;
+            let capacity = buffer.capacity();
+            let data = buffer.as_slice()?;
+
+            assert_eq!(data.len(), capacity, "Slice length should match capacity");
+            assert!(frame.size() <= capacity, "Frame size should be <= capacity");
+
+            // Verify we can read data without crashing
+            let _first_byte = data[0];
```
apljungquist: I like the idea and I think we should keep it. Maybe even add a variant for `as_mut_slice`. But I will still challenge you a little bit:

Are you sure these accesses aren't removed by the optimizer?
I seem to remember this being a pitfall when benchmarking: https://doc.rust-lang.org/stable/std/hint/fn.black_box.html

Also, if they are not removed, would they always fail if we cannot access the data or only if we are (un)lucky and the address points to a certain region of memory?

I think it may be worth documenting our answers to these questions, either with comments in the code or in the PR description which will later turn into the commit message.

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+            let _last_byte = data[frame.size().saturating_sub(1)];
+        }
+
+        running.stop()?;
+        Ok(())
+    }
+
+    #[test]
+    fn multiple_streams_sequential() -> std::result::Result<(), Box<dyn std::error::Error>> {
```
apljungquist: Nice test. Have you considered the use case of concurrent streams?

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+            // Drop should handle cleanup
+        }
+
+        // If we get here without crash, cleanup worked
+        log::info!("Stream dropped without explicit stop - cleanup successful");
+        Ok(())
+    }
+
+    #[test]
+    fn error_message_is_descriptive() {
```
apljungquist: How does this test (automatically) verify that the error message is descriptive?

apljungquist (commented):
I skimmed the PR and I think it looks pretty good. I'll try to do a more thorough read in the before next weekend focusing on safety and I'll try to respond to comments within 24h.

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+        if map.is_null() {
+            Err(Error::NullPointer)
+        } else {
+            Ok(Self { raw: map })
+        }
+    }
+
+    /// Sets a 32-bit unsigned integer value.
+    pub fn set_u32(&self, key: &str, value: u32) -> Result<()> {
+        let key_cstr = CString::new(key).map_err(|_| Error::CStringAllocation)?;
```
apljungquist: What do you think about taking the key as `&CStr`?

That would make this function infallible, relieving users of error handling.
In some cases creating the `CStr` would be annoying for the user, but in other cases they will be using `c"string"` literals anyway. It also gives the user the flexibility to decide how to handle conversion (errors), maybe they would rather do a lossy conversion.

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+    /// Default: 0 (main channel)
+    pub fn channel(mut self, channel: u32) -> Self {
+        self.channel = channel;
+        self
+    }
+
+    /// Sets the video resolution.
+    ///
+    /// If width or height is 0, the camera's native resolution is used.
+    pub fn resolution(mut self, width: u32, height: u32) -> Self {
```
apljungquist: What do you think about making width and height separate setter functions?
That way there is not ambiguity about which argument comes first.

It also makes it possible to make `0` an error. Or maybe we could have a `Resolution` enum with variants for `Native` and `Exact` (if that matches how the API works).

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+    pub fn iter(&mut self) -> StreamIterator<'_> {
+        StreamIterator {
+            stream: self.stream,
+        }
+    }
+
+    /// Stops the stream.
+    ///
+    /// After stopping, no more frames can be retrieved.
+    pub fn stop(&mut self) -> Result<()> {
```
apljungquist: What do you think about taking `self` instead of `&mut self`?

That way it will be clear when a stream is stopped (no `.stop` which don't do anything) and it will definitely be impossible to `iter` on a stopped stream.

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+            raw: frame,
+            _phantom: PhantomData,
+        })
+    }
+}
+
+impl Drop for StreamBuffer<'_> {
+    fn drop(&mut self) {
+        unsafe {
+            let _ = try_func!(vdo_stream_buffer_unref, self.stream.raw, &mut self.raw);
```
apljungquist: What do you think about adding an `unref` method to `StreamBuffer` that allows the user to obtain the error message?

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+        let data = unsafe { vdo_buffer_get_data(self.raw) };
+        if data.is_null() {
+            return Err(Error::NullPointer);
+        }
+        let slice = unsafe { std::slice::from_raw_parts_mut(data as *mut u8, self.capacity()) };
+        Ok(slice)
+    }
+
+    /// Returns frame metadata for this buffer.
+    pub fn frame(&self) -> Result<Frame<'_>> {
```
apljungquist: What do you think about returning an `Option` instead of a `Result`?

This maps quite nicely to a nullable pointer and users don't have to think about if/how to handle all the other error variants.

apljungquist (commented):
Some API design related questions.

API design is difficult and typically subjective. I don't want it to stall this PR, so consider this inspiration. (that's not to say that the previous review was not subjective or disputable, only that this is even more so)

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+}
+
+// ============================================================================
+// Frame - Frame metadata
+// ============================================================================
+
+/// Metadata for a video frame.
+///
+/// Contains information about frame timing, size, and type.
+pub struct Frame<'a> {
```
guoxe: Consider merging the `Frame` and `StreamBuffer` structs since their lifetimes are very much linked.
As stated in https://developer.axis.com/acap/api/src/api/vdostream/html/vdo-buffer_8h.html#a5367ff8edb99976f0b22a2809d44c097, the frame object is only valid while the buffer is so I'm not convinced it needs a separate type when we could just add relevant methods to the StreamBuffer type.

guoxe (changes requested):

crates/vdo/src/lib.rs
apljungquist: I'm not saying we should depend on this, but no shit they are linked: `typedef VdoFrame VdoBuffer;` :sweat_smile:

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+    /// Missing error data from VDO library.
+    #[error("Missing error data from VDO library")]
+    MissingVdoError,
+    /// No buffers are allocated for the stream.
+    #[error("No buffers are allocated for the stream")]
+    NoBuffersAllocated,
+}
+
+/// Result type for VDO operations.
+pub type Result<T> = std::result::Result<T, Error>;
```
apljungquist: nit; I would prefer it if we don't alias `Result` because writing out `Result<Error>` isn't that much work and if we have multiple result types it starts getting difficult understanding what a function returns by just reading the signature.

apljungquist (commented):
I have focused this review on the `Map` type, which looks good. Once we have decided if/how to change it I will continue reviewing other parts.

I do this to not overwhelm you with comments and to not overwhelm myself with different discussions to keep track of and large diffs that need to be reviewed with every now commit.

apljungquist:
Will try to get around to this tomorrow

crates/vdo/src/lib.rs
```diff
@@ -1315,45 +1245,69 @@ mod device_tests {
     fn error_message_is_descriptive() {
```
apljungquist: For tests that verify that an output is subjectively good for a human, I have found that `expect-test` is a useful tool. You don't have to, but if you feel like looking into it I think it could be useful in this test.

apljungquist:
@vsem-azamat nice progress!

Please refrain from force pushing in the future as it makes it difficult to review the changes and keep track of comments. I took the liberty of cherry-picking your fix onto the branch before it was force pushed and force pushing that, so now the git history looks as it should.

A few of my comments seem to have gone unaddressed. You don't have to always to things the way I think is the best, but I would appreciate if you at least acknowledge on each comment that you have read it and state what, if anything, you will do to address it :slightly_smiling_face: 

I haven't seen the results of the CI yet, but if we get more revisions in on this I will make sure to start it as soon as I can instead of waiting until I have time to do human review like this time around.

apljungquist (commented):
I forgot to submit this review, sorry. Among the new comments I only feel strongly about the glob import, but as I said, please respond to all new and old comments.

crates/vdo/src/map.rs
```diff
@@ -18,7 +20,7 @@ impl CStringPtr {
     /// allocated in a manner compatible with [`glib_sys::g_free`] and there must be no other
     /// users of this memory.
     pub(crate) unsafe fn from_ptr(ptr: *mut c_char) -> Self {
-        debug_assert!(!ptr.is_null());
+        assert!(!ptr.is_null(), "CStringPtr::from_ptr called with null");
         Self(NonNull::new_unchecked(ptr))
```
apljungquist: Why not use `NonNull::new`?

https://doc.rust-lang.org/beta/src/core/ptr/non_null.rs.html#270

crates/vdo/src/map.rs
```diff
@@ -62,26 +78,27 @@ impl Map {
     /// `ptr` must be a non-null, valid `VdoMap` pointer with ownership
```
apljungquist: > `ptr` must be a non-null,

This is no longer a SAFETY precondition, but a PANICS precondition

apljungquist (commented):
Some nice improvements in this iteration. But there are still old comments that remain unaddressed. You may want to also take at least a quick look at Gustaf's comments as he he the person who will ultimately merge your PR if approved 😉

crates/vdo/src/lib.rs
```diff
@@ -461,54 +467,65 @@ impl StreamBuffer<'_> {
...
+            None
+        } else {
+            Some(size as usize)
+        }
     }
 
     /// Returns a borrowed file descriptor for the buffer's backing memory.
+    ///
+    /// The fd is only valid for the lifetime of this buffer. Do not convert it to
+    /// an `OwnedFd` (e.g. via `try_clone_to_owned`), as the underlying fd is closed
```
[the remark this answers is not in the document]
vsem-azamat: Good catch. Any safe API that allows users to trigger undefined behavior is unsound and must be marked unsafe.
Changed the return type from `BorrowedFd<'_>` to `RawFd` and marked the method `unsafe`.

crates/vdo/src/map.rs
vsem-azamat: I'll switch to `NonNull::new(...).expect(...)`.

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+        let message = if g_error.message.is_null() {
+            String::from("Unknown error")
+        } else {
+            unsafe { CStr::from_ptr(g_error.message) }
+                .to_str()
+                .unwrap_or("Invalid UTF-8 in error message")
+                .to_string()
+        };
+
+        // Free the GError
```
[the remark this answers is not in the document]
vsem-azamat: Done - added `// SAFETY:` comments on unsafe blocks, e.g. [here](https://github.com/vsem-azamat/acap-rs/blob/feature/vdo-completion/crates/vdo/src/lib.rs#L110-L112).

crates/vdo/src/lib.rs
vsem-azamat: `Frame` was merged into [`StreamBuffer`](https://github.com/vsem-azamat/acap-rs/blob/feature/vdo-completion/crates/vdo/src/lib.rs#L438). 
All frame metadata methods are now on StreamBuffer directly.

apljungquist (dismissed):

apljungquist (dismissed):

guoxe (commented):
I have some questions relating to some of the functions added.
Overall I think it looks good!

crates/vdo/src/map.rs
```diff
@@ -0,0 +1,170 @@
...
+/// All methods assume `self.raw` is a valid `VdoMap` pointer, which is
+/// guaranteed by the constructors.
+pub struct Map {
+    raw: *mut VdoMap,
+}
+
+impl Map {
+    pub fn try_new() -> std::result::Result<Self, super::Error> {
+        let map = unsafe { vdo_sys::vdo_map_new() };
+        if map.is_null() {
```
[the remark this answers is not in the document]
vsem-azamat: Right, the null path was unreachable. 

`Map::new()` is now infallible and returns `Self` directly. 
Fixed in 7489918.

crates/vdo/src/map.rs
```diff
@@ -0,0 +1,170 @@
...
+            unsafe { vdo_sys::vdo_map_dup_string(self.raw, key.as_ptr(), ptr::null::<c_char>()) };
+        if ptr.is_null() {
+            return None;
+        }
+        // SAFETY: ptr is non-null, allocated by g_malloc via vdo_map_dup_string, and we own it.
+        Some(unsafe { CStringPtr::from_ptr(ptr) })
+    }
+
+    pub fn set_bool(&mut self, key: &CStr, value: bool) {
+        let gvalue = if value {
```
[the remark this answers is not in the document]
vsem-azamat: Done: `set_bool/get_bool` use `IntoGlib/from_glib` from the glib crate. 

Fixed in 7489918.

crates/vdo/src/map.rs
```diff
@@ -0,0 +1,170 @@
...
+            glib_sys::GTRUE
+        } else {
+            glib_sys::GFALSE
+        };
+        unsafe { vdo_sys::vdo_map_set_boolean(self.raw, key.as_ptr(), gvalue) }
+    }
+
+    pub fn get_bool(&self, key: &CStr, default: bool) -> bool {
+        let gdefault = if default {
```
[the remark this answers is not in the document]
vsem-azamat: Same fix, see above

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1330 @@
...
+    }
+
+    pub fn is_last_buffer(&self) -> bool {
+        unsafe { vdo_sys::vdo_frame_get_is_last_buffer(self.raw) != glib_sys::GFALSE }
+    }
+
+    /// Explicitly unreferences this buffer, returning an error if the operation fails.
+    ///
+    /// Normally buffers are unreferenced on drop. Use this if you need error handling.
+    pub fn unref(self) -> std::result::Result<(), Error> {
```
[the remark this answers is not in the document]
vsem-azamat: Removed, Drop handles cleanup: 7489918

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1330 @@
...
+    raw: *mut VdoStream,
+    started: bool,
+}
+
+// SAFETY: We hold exclusive ownership of the raw pointer and the VDO SDK
+// does not require streams to be pinned to a specific thread.
+unsafe impl Send for Stream {}
+
+impl Stream {
+    pub fn builder() -> StreamBuilder {
```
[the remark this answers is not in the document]
vsem-azamat: Removed; `StreamBuilder::new()` is now the only entry point.

crates/vdo/src/lib.rs
```diff
@@ -451,426 +301,350 @@ impl StreamBuilder {
...
+        // SAFETY: VDO buffers are backed by mmap'd or allocated regions that are fully
+        // initialized at allocation time. Bytes beyond size() may be stale but are valid.
         let slice = unsafe { std::slice::from_raw_parts(data as *const u8, self.capacity()) };
         Ok(slice)
     }
 
-    /// Returns the frame data as a mutable byte slice.
-    pub fn as_mut_slice(&mut self) -> Result<&mut [u8]> {
-        let data = unsafe { vdo_buffer_get_data(self.raw) };
+    /// Returns a copy of exactly [`size()`](StreamBuffer::size) bytes of frame data.
```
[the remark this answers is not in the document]
vsem-azamat: Fixed: now it offsets by `header_size()` and the doc comment states that the header is excluded. `as_slice()` remains for a raw view of the whole buffer.

crates/vdo/src/lib.rs
```diff
@@ -451,426 +301,350 @@ impl StreamBuilder {
...
+    /// # Safety
     ///
-    /// This can be used for zero-copy operations with other APIs.
-    pub fn file_descriptor(&self) -> std::os::fd::BorrowedFd<'_> {
-        unsafe {
-            let fd = vdo_frame_get_fd(self.raw);
-            std::os::fd::BorrowedFd::borrow_raw(fd)
+    /// The returned fd is owned by VDO and will be closed when this buffer is
+    /// unreferenced. The caller must not close or duplicate (`dup`) the fd.
+    pub unsafe fn file_descriptor(&self) -> std::result::Result<std::os::fd::RawFd, Error> {
```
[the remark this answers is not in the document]
vsem-azamat: I've removed `file_descriptor`. You're right, I'll add it when I test it (most likely in #230 ) and understand if it's really needed. 

Removed in 17f0d26.

vsem-azamat:
Hi!
@guoxe 
I've brought some updates :smile:

crates/vdo/src/lib.rs
```diff
@@ -0,0 +1,1359 @@
...
+    }
+
+    /// Returns the header size in bytes.
+    pub fn header_size(&self) -> isize {
+        unsafe { vdo_frame_get_header_size(self.raw) }
+    }
+
+    /// Returns the file descriptor for the frame data.
+    ///
+    /// This can be used for zero-copy operations with other APIs.
```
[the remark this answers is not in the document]
guoxe: Acknowledged

guoxe (dismissed):
Nice work!

Sorry it took a while for me to get around to reviewing, unless there is something else you want to address I am happy with merging this PR as is.

vsem-azamat:
@guoxe 
I've fixed checksum. Please, approve it again

guoxe (approved):

vsem-azamat:
> @guoxe I've fixed checksum. Please, approve it again

@apljungquist @guoxe 

Thanks you so much for for helping and reviewing this PR. It was tough for me.

I'd like to merge this PR and in anything appears I'll open new FollowUp MR. 

Thanks)

apljungquist:
I'm happy you were up for the challenge and that it hasn't discouraged you :)

PR #224, opened 2026-01-07
chore: Document ACAP Native API bindings priorities

apljungquist:
What makes a change an improvement is a question I have struggled with both as an author and as a reviewer. Coming up with an answer starts with deciding what the goal of a project is. My hope in writing down these priorities is that it will save others the time it would take to rediscover them and that it will make them stable across reviewers and time (I have changed my mind more than once).

### DIFF
--- a/docs/project.md
+++ b/docs/project.md
@@ -14,7 +14,7 @@ The quality of a library has many facets:
   - etc
 - Flexibility: What portion of the programs that one may want to express with the underlying C API can be expressed with the Rust bindings.
 - Performance: How much storage, memory and CPU does an application built using the bindings use.
-- Robustness: Can non-breaking changes to the underlying C API be made without breaking the Rust bindings and could such.
+- Robustness: Can non-breaking changes to the underlying C API be made without breaking the Rust bindings.
 - Soundness: Safe Rust using the library cannot cause undefined behavior [^1].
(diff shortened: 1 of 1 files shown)


docs/project.md
```diff
@@ -1,3 +1,38 @@
...
+
+- Ergonomics:
+  - How straightforward to use are the bindings
+  - How readable are applications using the bindings
+  - How fast do the bindings compile
+  - How much can be achieved without `unsafe`
+  - etc
+- Flexibility: What portion of the programs that one may want to express with the underlying C API can be expressed with the Rust bindings.
+- Performance: How much storage, memory and CPU does an application built using the bindings use.
+- Robustness: Can non-breaking changes to the underlying C API be made without breaking the Rust bindings and could such.
```
guoxe: Is a part missing from this one? "and could such" seems like the start of a sentence.

guoxe:
> @guoxe what do you think about writing down something like this?
> 
> I think it would have helped in deciding how to handle #207 (I think I ended up making the wrong decision) and I hope that it will help when reviewing future PRs

I think it looks good, can't hurt to have it written down!

docs/project.md
apljungquist: Oops. The second part was the start of what later became the stability facet.

PR #239, opened 2026-07-09
chore: Replace `acap-build` crate with external `rs4a-eap`

apljungquist:
> [!NOTE]
> This is still work in progress, I still want to land a few more changes in rs4a and publish the crate.
> That said, early feedback is welcome, in particular on issues such as:
> - whether the factoring out is at all acceptable
> - pushing environment variable parsing into the CLI; this is kind of a drive by change but it's a practice that has grown on me since it keeps library functions pure and makes it easy for users of a CLI to see all inputs that affect the behavior of the program

Remove the in-tree `acap-build` crate and depend on the `rs4a-eap` crate from apljungquist/rs4acap instead. This has benefits in two categories:
- Clarifies the purpose of this repo as facilitating the use of Rust in ACAP app implementations
- `rs4a-eap` has several implemented and planned improvements, notably the option to not depend on external programs allowing it to be fully installed with cargo, and more robust testing.

We drop all use of the reference implementation, which rs4a does not support and instead rely on the testing done by the library maintainer (me). The approach was flawed anyway since bugs could creep in when staging files in a temporary directory.

Footers
=======

Related-to: #205 (abandoned attempt at not shelling out to `cp`, a problem that does not affect `rs4a-eap`)


crates/cargo-acap-build/src/main.rs
```diff
@@ -68,7 +78,7 @@ fn build_and_copy(cli: Cli) -> anyhow::Result<()> {
     AppBuilder::from_targets(cli.targets())
         .args(args)
         .artifact_dir(get_cargo_metadata(None)?.target_directory.join("acap"))
-        .execute()?;
+        .execute(cli.source_date_epoch.unwrap_or_default())?;
```
apljungquist, line 85: This doesn't actually match the docstring:

> Defaults to the current time.

I need to fix that.

I may move the interpolation into execute or some other shared location.

crates/cargo-acap-sdk/src/main.rs
```diff
@@ -69,12 +70,24 @@ enum Commands {
     Completions(CompletionsCommand),
 }
 
+fn parse_mtime(s: &str) -> anyhow::Result<Mtime> {
```
guoxe, line 73: Might be worth DRYing this since to me it does not seem like we would want the implementations to diverge.

guoxe (commented):
I'm not against this change. Especially as more parts of the ACAP SDK gets added it would be nice to focus on only the wrappers and example applications and having any additional tooling live elsewhere.
This way, the focus of this repo is clearer and there is really nothing that requires one to use the extra tooling either, using cargo and existing ACAP tools works just as well as the experimental ones.

PR #241, opened 2026-07-11
feat: Add sys-crate for the Device Data Hub API

apljungquist:
Device Data Hub is the designated replacement for the Message Broker API, which is deprecated and will be removed in AXIS OS 13.0.

The bindings are generated from the C headers the same way as for `mdb-sys`. Constants that the C headers define as object-like macros with casts are transcribed manually because `bindgen` cannot evaluate them; we may be able to remove those if we upgrade to `bindgen >= 0.70.0`

Details
=======

`.devcontainer/Dockerfile`:
- Set `AXIS_OS_VERSION` or the new example won't build.

`apps/object_consumer`:
- Port an example to trigger the sys-crate build script or the `crates/%-sys/src/bindings.rs` make target fails [^1]. It will be updated to use the safe bindings when they are added.

[^1]: https://github.com/AxisCommunications/acap-rs/actions/runs/30839862786/job/91773991360#step:5:381

Footers
=======

Depends-on: #240


apljungquist:
Looks like this is failing because the two `acap-build` implementations have diverged. Parking this until #239 is done.

apljungquist:
> Looks like this is failing because the two `acap-build` implementations have diverged. Parking this until #239 is done.

I forgot about this

PR #248, opened 2026-07-18
fix(vdo): compile against firmware-11 SDK 1.15.1 (VDO_ERROR_NO_VIDEO)

### LINKED ISSUE #245: `vdo` does not compile against firmware-11 Native SDK 1.15.1 (`VDO_ERROR_NO_VIDEO`)
## Summary

The safe `vdo` crate does not compile against the **firmware-11 Native SDK (`1.15.1`)** sysroot, because `VdoError::code_name()` references `vdo_sys::VDO_ERROR_NO_VIDEO`, which does not exist in the 1.15.1 VDO headers (it was added in a later SDK). Building fails with:

```
error[E0425]: cannot find value `VDO_ERROR_NO_VIDEO` in crate `vdo_sys`
   --> crates/vdo/src/lib.rs:148:32
    |
148 |             x if x == vdo_sys::VDO_ERROR_NO_VIDEO.0 as i32 => "VDO_ERROR_NO_VIDEO",
    |                                ^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `VDO_ERROR_NO_DATA`
error: could not compile `vdo` (lib) due to 1 previous error
```

`vdo-sys`'s bindgen output for the 1.15.1 sysroot simply doesn't define that constant, so the reference in the higher-level crate is unconditional and breaks the build.

## Repro

Cross-compile any consumer of `vdo` with the firmware-11 SDK image, e.g.:

```
axisecp/acap-native-sdk:1.15.1-armv7hf-ubuntu22.04
axisecp/acap-native-sdk:1.15.1-aarch64-ubuntu22.04
```

`vdo-sys` builds fine (bindings generate against the 1.15.1 sysroot); only the safe `vdo` crate fails, at the `code_name()` match arm above.

## Why it matters

Firmware-11 is the current LTS and covers a large installed base (e.g. ARTPEC-6/7 cameras that don't take firmware 12). Being able to target it from Rust is valuable; today the `vdo` crate is effectively firmware-12-only.

## Suggested direction

`code_name()` is a diagnostic error-code → string map with a `_ => "VDO_ERROR_UNKNOWN"` catch-all, so the missing arm is not load-bearing. Options, in rough order of preference:

1. Gate SDK-version-specific arms behind a `cfg`/build-probe so the crate compiles across the SDK range it supports.
2. Or drop the `VDO_ERROR_NO_VIDEO` arm (the fallback already covers the code) if a single lowest-common-denominator set is acceptable.

This is the inverse of #244 (which tracks *newer* SDK symbols) — the same "one `vdo` source, multiple SDK versions" portability question, but for an *older* SDK than the one the bindings were generated against.

## Context / offer

Found while building a real ACAP app (on-camera LL-HLS via VDO) for an ARTPEC-6 / firmware-11.11 camera. Verified end-to-end after applying option (2) in a fork. Happy to open a PR for whichever approach you'd prefer.

fishloa:
Fixes #245.

`VdoError::code_name()` referenced `vdo_sys::VDO_ERROR_NO_VIDEO`, which the **firmware-11 Native SDK (`1.15.1`)** vdo headers don't define (it was added in a later SDK), so the `vdo` crate failed to build against that sysroot:

```
error[E0425]: cannot find value `VDO_ERROR_NO_VIDEO` in crate `vdo_sys`
   --> crates/vdo/src/lib.rs:148:32
```

The arm is diagnostic-only and the `_ => "VDO_ERROR_UNKNOWN"` catch-all already covers the code, so dropping it lets `vdo` compile against **both** the fw11 (`1.15.1`) and fw12 (`12.1.0`) SDK sysroots.

Verified end-to-end: with this change the crate cross-compiles under `axisecp/acap-native-sdk:1.15.1-{aarch64,armv7hf}-ubuntu22.04` and a real ACAP app (on-camera LL-HLS via VDO) runs on an ARTPEC-6 / firmware-11.11 camera.

If you'd rather **keep** the label for fw12+ instead of dropping it, a `cfg`/build-probe gate would also work — happy to switch this PR to that approach.

### DIFF
--- a/crates/vdo/src/lib.rs
+++ b/crates/vdo/src/lib.rs
@@ -145,7 +145,6 @@ impl VdoError {
             x if x == vdo_sys::VDO_ERROR_FATAL.0 as i32 => "VDO_ERROR_FATAL",
             x if x == vdo_sys::VDO_ERROR_NOT_CONTROLLED.0 as i32 => "VDO_ERROR_NOT_CONTROLLED",
             x if x == vdo_sys::VDO_ERROR_NO_EVENT.0 as i32 => "VDO_ERROR_NO_EVENT",
-            x if x == vdo_sys::VDO_ERROR_NO_VIDEO.0 as i32 => "VDO_ERROR_NO_VIDEO",
             _ => "VDO_ERROR_UNKNOWN",
         }
     }


apljungquist:
I think we should find a way to compile conditionally with the error code because I don't think the bindings will be particularly useful if we reduce them to supporting only the greatest common denominator. I haven't thought much about exactly how to do this, so I'm happy to discuss options or just go with whatever you propose as long as it sounds reasonable (with the reservation that we may change it down the road if we come up with something that has tangible benefits). But I would like to use this as a starting point (feel free to challenge me if you disagree): We should support the latest minor/micro version within each major version of AXIS OS.

apljungquist (changes requested):
Changing the verdict to _request changes_ to make the status of the PR clearer, see earlier comment for details.

PR #249, opened 2026-07-18
feat(vdo): add StreamBuilder::gop_length

### LINKED ISSUE #246: `StreamBuilder`: expose GOP length (+ `force_key_frame`) for key-frame control
## Summary

`StreamBuilder` exposes `channel` / `format` / `resolution` / `framerate` / `buffers`, but no way to set the **GOP length** (key-frame interval). On cameras running dynamic-GOP / Zipstream, key frames can be many seconds (or scene-changes) apart, which makes a consumer that needs to observe a key frame at startup — e.g. to recover SPS/PPS, or to obtain a clean random-access point for CMAF/HLS segmenting — wait an unbounded amount of time or give up.

## Repro / observed

On an ARTPEC-6 / firmware-11 camera in its default (dynamic-GOP) config, an H.264 stream delivered **one** key frame at subscribe then only P-frames for 150+ buffers (>5 s) — no further key frame in a bounded scan window.

## Request

Add an optional GOP-length setter to `StreamBuilder`, passed through to `vdo_stream_new` as the `gop_length` setting, e.g.:

```rust
let stream = StreamBuilder::new()
    .format(VdoFormat::VDO_FORMAT_H264)
    .framerate(30)
    .gop_length(30) // ~1 s; 0 / unset = camera default
    .build()?;
```

`vdo_stream_force_key_frame` (already in `vdo-sys`) would also be a useful safe wrapper on the running stream for the same use case.

## Context / offer

Implemented `StreamBuilder::gop_length` (additive, `0` = camera default, sets `c"gop_length"` in `build()`'s `VdoMap`) in a fork and it works on hardware. Glad to send a PR if you're open to it.

fishloa:
Fixes #246.

Adds `StreamBuilder::gop_length(u32)`, passed through to `vdo_stream_new` as the `gop_length` setting (`0`/unset = camera default), so callers can pin the key-frame interval:

```rust
let stream = StreamBuilder::new()
    .format(VdoFormat::VDO_FORMAT_H264)
    .framerate(30)
    .gop_length(30) // ~1 s
    .build()?;
```

Motivation: on cameras in dynamic-GOP / Zipstream mode, key frames can be many seconds apart, so a consumer that must observe a key frame at startup — to recover SPS/PPS, or to get a clean CMAF/HLS random-access point — can otherwise wait unbounded. Observed on an ARTPEC-6 / firmware-11 camera: one key frame at subscribe then only P-frames for 150+ buffers.

Additive and backward-compatible (new builder field defaults to `0`; the setting is only written when non-zero). A safe wrapper for `vdo_stream_force_key_frame` would complement this — happy to add it here or in a follow-up if wanted.


apljungquist (dismissed):
LGTM. Ideally the PR description / commit message should document why we don't need the ability to pass `gop_length=0` to the C implementation to make it easier for future contributors to understand when this behavior can or cannot be changed.

apljungquist (approved):

apljungquist:
@guoxe I think my approval is insufficient to enable merging, could you take a quick look?

@vsem-azamat @deepikas20 is it ok if I ask you for final approval on changes I think look good, such as this one (assuming your approvals are sufficient, I don't know if the repo is set up to require repo-admin approval or just organization-member approval)?

guoxe (approved):

PR #250, opened 2026-07-18
docs(vdo): key-frame SPS/PPS live in the stripped header; add header_bytes()

### LINKED ISSUE #247: `data_copy()` silently strips the key-frame header carrying SPS/PPS — document / add accessor
## Summary

`StreamBuffer::data_copy()` strips `header_size()` bytes off the front of the frame ("excluding the header if one is present"). For **H.264/H.265 key frames**, that stripped header is where VDO carries the **SPS/PPS (and VPS) parameter sets** — so `data_copy()` on a key frame returns only the coded slice, *without* the parameter sets, and there is nothing in the current docs to signal this.

This is a genuine footgun: a caller that scans `data_copy()` bytes for SPS/PPS (to build an `avcC`/`hvcC`, an MP4 init segment, an SDP fmtp, etc.) will never find them, even though the stream is perfectly healthy.

## Evidence

On an ARTPEC-6 / firmware-11 H.264 stream, for the IDR buffer:

```
frame_type = VDO_FRAME_TYPE_H264_IDR
header_size = Some(41)
as_slice()[..size()] head = 00 00 00 01 67 64 00 29 ...   // 0x67 = SPS NAL, then PPS, then the IDR slice
data_copy()         head = 00 00 00 01 65 ...              // 0x65 = IDR slice only — SPS/PPS gone
```

So the parameter sets live entirely within the `[0 .. header_size)` region that `data_copy()` discards; `as_slice()[..size()]` is the full `[SPS][PPS][IDR slice]` Annex-B access unit.

## Request

1. **Docs:** on `data_copy()` (and/or `header_size()`), note that on key frames the header contains the SPS/PPS/VPS parameter sets, and that `as_slice()` is the way to obtain them.
2. **Optional API:** a small convenience like `parameter_sets()` / `header_bytes()` returning the `[0 .. header_size)` slice would make the common "grab the codec config once" path obvious and hard to get wrong.

## Context

Cost a few debugging cycles on real hardware (the stream looked fine, key frames arrived, but SPS/PPS never appeared) before `header_size` made it obvious. Filing so the next person doesn't repeat it. Happy to PR the doc note + accessor.

fishloa:
Fixes #247.

`data_copy()` strips the `[0, header_size())` header, and on H.264/H.265 **key frames** that header is exactly where VDO carries the SPS/PPS (and VPS) parameter sets — so `data_copy()` on a key frame returns only the coded slice, and a caller scanning those bytes for parameter sets never finds them even on a perfectly healthy stream.

Observed on an ARTPEC-6 / firmware-11 H.264 IDR buffer:

```
header_size        = Some(41)
as_slice()[..size] = 00 00 00 01 67 64 00 29 ...   // 0x67 = SPS, then PPS, then the IDR slice
data_copy()        = 00 00 00 01 65 ...            // 0x65 = IDR slice only — SPS/PPS gone
```

This PR:
- documents the behaviour on `data_copy()` and `header_size()`, and
- adds `StreamBuffer::header_bytes()` returning the `[0, header_size())` slice, so the "grab the codec config once" path is obvious and hard to get wrong.

No behaviour change to existing methods; `header_bytes()` mirrors `data_copy()`'s existing unsafe access pattern.

### DIFF
--- a/crates/vdo/src/lib.rs
+++ b/crates/vdo/src/lib.rs
@@ -453,9 +453,17 @@ impl StreamBuffer<'_> {
         Ok(slice)
     }
 
-    /// Returns a copy of the frame data, excluding the header if one is present.
+    /// Returns a copy of the frame data, excluding the header if one is present
+    /// (the `[0, header_size())` prefix — see [`header_bytes()`](StreamBuffer::header_bytes)).
     ///
-    /// Use [`as_slice()`](StreamBuffer::as_slice) for a raw view of the whole buffer.
+    /// **On H.264/H.265 key frames this excludes the codec parameter sets.**
+    /// VDO carries the SPS/PPS (and VPS for H.265) in that stripped header, so
+    /// `data_copy()` on a key frame returns only the coded slice — the
+    /// parameter sets are *not* in the result. To recover them (e.g. to build
+    /// an `avcC`/`hvcC`, an MP4 init segment, or an SDP `fmtp`), read
+    /// [`header_bytes()`](StreamBuffer::header_bytes), or use
+    /// [`as_slice()`](StreamBuffer::as_slice) for the full `[header][data]`
+    /// buffer.
     pub fn data_copy(&self) -> std::result::Result<Vec<u8>, Error> {
         let data = unsafe { vdo_sys::vdo_buffer_get_data(self.raw) };
         if data.is_null() {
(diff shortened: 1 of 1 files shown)


crates/vdo/src/lib.rs
```diff
@@ -475,6 +483,32 @@ impl StreamBuffer<'_> {
...
+    /// [`data_copy()`](StreamBuffer::data_copy) excludes — or an empty slice
+    /// when the frame has no header.
+    ///
+    /// On H.264/H.265 **key frames** this header carries the codec parameter
+    /// sets (SPS/PPS, and VPS for H.265) in Annex B, so this is where to read
+    /// them to build an `avcC`/`hvcC`, an MP4 init segment, or an SDP `fmtp`.
+    /// The full access unit is `header_bytes()` followed by `data_copy()`
+    /// (equivalently, [`as_slice()`](StreamBuffer::as_slice) up to
+    /// [`size()`](StreamBuffer::size)).
+    pub fn header_bytes(&self) -> std::result::Result<&[u8], Error> {
```
apljungquist, line 496: I haven't looked at what interpreting these bytes looks like, but I am wondering if perhaps there is an opportunity here to encode some more information about the header?
i.e. what would it look like if we returned something like `Result<Option<Header>, Error>`

crates/vdo/src/lib.rs
```diff
@@ -453,9 +453,17 @@ impl StreamBuffer<'_> {
...
-    /// Use [`as_slice()`](StreamBuffer::as_slice) for a raw view of the whole buffer.
+    /// **On H.264/H.265 key frames this excludes the codec parameter sets.**
+    /// VDO carries the SPS/PPS (and VPS for H.265) in that stripped header, so
+    /// `data_copy()` on a key frame returns only the coded slice — the
+    /// parameter sets are *not* in the result. To recover them (e.g. to build
+    /// an `avcC`/`hvcC`, an MP4 init segment, or an SDP `fmtp`), read
+    /// [`header_bytes()`](StreamBuffer::header_bytes), or use
+    /// [`as_slice()`](StreamBuffer::as_slice) for the full `[header][data]`
+    /// buffer.
     pub fn data_copy(&self) -> std::result::Result<Vec<u8>, Error> {
```
apljungquist, line 467: Maybe we should replace `data_copy` with `data_bytes` for consistency and flexibility in a follow up commit?

Adding `.to_vec()` to call sites is a minor inconvenience.

crates/vdo/src/lib.rs
```diff
@@ -475,6 +483,32 @@ impl StreamBuffer<'_> {
         Ok(slice.to_vec())
     }
 
+    /// Returns the frame's header bytes — the `[0, header_size())` prefix that
```
apljungquist, line 486: > `[0, header_size())`

Personally I would leave out details about where in the buffer the header bytes are from the commit message

apljungquist (approved):
LGTM

crates/vdo/src/lib.rs
apljungquist, line 467: You expressed surprise that _data_ does not include the _header_ and I'm thinking maybe we can help future users by choosing better names for the concept?

Having two functions with similar names e.g. `header_bytes` and `data_bytes` would also help I imagine.

apljungquist (commented):
I like the context in the issue btw; I didn't read the issue at first because it looked like a wall of AI text but now that I have read it that section helps convince me that there is a real developer with a real use case behind the isse+PR 😄 

> Cost a few debugging cycles on real hardware (the stream looked fine, key frames arrived, but SPS/PPS never appeared) before header_size made it obvious. Filing so the next person doesn't repeat it. Happy to PR the doc note + accessor.

guoxe (approved):

PR #252, opened 2026-08-06
fix(cargo-acap-build): Set `-mfpu=neon` for armv7hf

### LINKED ISSUE #251: Possible target triple issue when building for Cortex A9 products
### Describe the bug

Compilation target "thumbv7neon-unknown-linux-gnueabihf" may not work for all Axis products.

When building with target "thumbv7neon-..." the CC crate will automatically assume vfpv4 which is not available on IMX6SX, Cortex A9, f.ex. This caused an issue when compiling rusqlite with bundled sqlite and adding a table with `updated_at datetime default(datetime('subsec')),`. The resulting ACAP, when run a A9210, would crash with a SIGILL as soon as an insert was made to that table.

### To reproduce

1. Build an ACAP that uses rusqlite, or anything else that produces vfpv4 instructions. If using rusqlite, use the "bundled" feature.
2. Insert a row into the table that has the `updated_at datetime default(datetime('subsec')),` row.
3. Observe the crash when running

### Environment

- Axis device model: A9210 but most devices with Cortex A9 should have the issue.
- Axis device firmware version: 12.
- Version: SDK 12.10

apljungquist:
I don't fully understand how to determine the correct options, but I'm told this should work and it does avoid the crash on my reproducer.

Fixes: #251


guoxe (approved):
Seems correct according to https://gcc.gnu.org/onlinedocs/gcc-16.1.0/gcc/ARM-Options.html .

I don't know if we should set -mthumb as well for the linker, have not experimented with that. This is what the environment-setup script of the SDK sets:
`-mthumb -mfpu=neon -mfloat-abi=hard -mcpu=cortex-a9`

apljungquist:
> Seems correct according to https://gcc.gnu.org/onlinedocs/gcc-16.1.0/gcc/ARM-Options.html .
> 
> I don't know if we should set -mthumb as well for the linker, have not experimented with that. This is what the environment-setup script of the SDK sets: `-mthumb -mfpu=neon -mfloat-abi=hard -mcpu=cortex-a9`

I was thinking about that too as it would be annoying to run into this kind of issue again. But I don't fully understand it, so I prefer to take small steps for now.

PR #253, opened 2026-08-07
feat!: Add partial support for armv7-unknown-linux-gnueabihf

apljungquist:
That the build tools force the user to compile for `thumbv7neon` when `armv7` would also work has bothered me for a while, but it has not been a problem before #251 where building for `armv7` would probably have been a quick workaround. This commit makes it possible to build for `armv7` using `cargo-acap-build`.

Details
=======

`crates/acap-build/src/lib.rs`:
- Allowed targets are still modelled as a closed set because to make sure all inputs map to a package architecture correctly and that the CLI documents and fails early where such a mapping is not implemented.
- The default target remains `thumbv7neon` because it is not incorrect and whatever the reasons for choosing that as our preferred target were I don't think they have meaningfully changed.

`crates/cargo-acap-build/src/main.rs`:
- Continue accepting the architecture nicknames as arguments because they are easier to remember and type than the full cargo target triple.
- Use the `--target` option for the target triple to make the interface familiar to rust programmers; `--arch` does to my knowledge not collide with any upstream options.

`crates/cargo-acap-sdk/src/main.rs`:
- Rename `--target` to `--arch` to align with `cargo` and `cargo-acap-build`.
- `From<ArchAbi> for Target` is not implemented because it would implie a one-to-one mapping which does not exist.

`crates/cargo-acap-build/src/main.rs`,
`crates/cargo-acap-sdk/src/main.rs`:
- Use the shortened "arch" instead of "architecture" because the latter is a lot to type.


.devcontainer/Dockerfile
guoxe, line 1: Since this PR adds armv7 as a target, maybe we could also add the other options to the thumb target suggested in https://github.com/AxisCommunications/acap-rs/pull/252#pullrequestreview-4880869334 ?

crates/cargo-acap-build/src/cargo_acap.rs
```diff
@@ -74,12 +73,12 @@ pub fn build_and_pack(
                     artifacts.push(Artifact::Eap {
                         path: pack(
                             &cargo_target_directory,
-                            arch,
+                            Architecture::from(target),
```
guoxe: Could use `target.into()` here for a more terse option.

crates/cargo-acap-build/src/lib.rs
```diff
@@ -17,24 +19,44 @@ mod command_utils;
...
-        U: Into<Architecture>,
+        U: Into<Target>,
     {
-        Self {
-            targets: targets.into_iter().map(|t| t.into()).collect(),
+        let targets: Vec<Target> = targets.into_iter().map(|t| t.into()).collect();
+
+        let mut by_architecture: HashMap<Architecture, Target> = HashMap::new();
+        for &target in &targets {
+            let architecture = Architecture::from(target);
```
guoxe: Can use `.into()` here as well if desired.

Makefile
```diff
@@ -93,7 +93,7 @@ stop:
 ## * The device has SSH enabled the ssh user root configured.
 run:
 	CARGO_TARGET_DIR=target-$(AXIS_DEVICE_ARCH) \
-	cargo-acap-build --target $(AXIS_DEVICE_ARCH) -- -p $(AXIS_PACKAGE) --profile dev
+	cargo-acap-build --arch $(AXIS_DEVICE_ARCH) -- -p $(AXIS_PACKAGE) --profile dev
```
guoxe, line 96: I think check_build target may have to be updated as well? When I tested it locally it did not work.

`➜  acap-rs git:(cargo-targets) ✗ make check_build
CARGO_TARGET_DIR=target-aarch64 \
cargo-acap-build \
	--arch aarch64 \
	-- \
	--package '*_*' \
	--profile dev \
	--locked
error: unexpected argument '--arch' found

  tip: to pass '--arch' as a value, use '-- --arch'

Usage: cargo-acap-build [OPTIONS] [ARGS]...

For more information, try '--help'.
make: *** [Makefile:280: target-aarch64/acap/_envoy] Error 2`

guoxe (commented):
Overall I think this is a good idea, can't hurt to allow armv7 as well.
IIRC the original reason for picking thumb as a default is because the instructions are smaller which should result in smaller binaries.

.devcontainer/Dockerfile
apljungquist: I have given this some more consideration and I don't think it's necessary; the reason for issues/251 and the resulting pulls/252 is that cc passes the wrong arguments to the C compiler. Our configuration was perfectly OK (I'm proposing we revert the workaround in #254).

Furthermore, this is what an AI tells me about the remaining flags:

| SDK flag                                             | Verdict                                                                                                                                                            |
| ---------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `-mfloat-abi=hard`                                   | Redundant — `arm-linux-gnueabihf-gcc` and the `gnueabihf` triple are hard-float already.                                                                           |
| `-mcpu=cortex-a9`                                    | Tuning/scheduling only. Not correctness once FPU is right; would pin to A9 and forfeit tuning on newer armv7hf parts.                                              |
| `-mthumb`                                            | Cosmetic/size; interworking is handled. Not needed.                                                                                                                |

Even if `-mthumb` is cosmetic, cc probably sets it [^2] though I haven't verified this empirically.

Related question, what part of the docs did you read to draw this conclusion? 

> Seems correct according to https://gcc.gnu.org/onlinedocs/gcc-16.1.0/gcc/ARM-Options.html . [^1]

[^1]: https://github.com/AxisCommunications/acap-rs/pull/252#pullrequestreview-4880869334
[^2]: https://github.com/rust-lang/cc-rs/blob/a91e05ec40f26d4637d4bff9e9764221d0a59dd8/src/lib.rs#L2545

Makefile
apljungquist: Did you update `cargo-acap-build` before running `make check_build`?

crates/cargo-acap-build/src/cargo_acap.rs
apljungquist: Done

crates/cargo-acap-build/src/lib.rs
apljungquist: Done.

Makefile
guoxe: I did not, that fixed it!

guoxe (approved):

PR #254, opened 2026-08-08
fix: Update cc and drop CFLAGS workaround

apljungquist:
The problem reported in #251 seems to be caused by `cc` passed `-mfpu=neon-vpfv4` the C compiler instead of `-mfpu=neon`, which is wrong [^1] [^2]. This is fixed in 1.4.2, so we can already update `cc` and drop our workaround again.

[^1]: https://github.com/rust-lang/cc-rs/issues/1842#issuecomment-5217445440
[^2]: rustc +nightly -Z unstable-options --print target-spec-json --target thumbv7neon-unknown-linux-gnueabihf | jq '.features' prints
"+v7,+thumb-mode,+thumb2,+vfp3,+neon"


.devcontainer/Dockerfile
apljungquist, line 1: Note to self: I need to test that this actually works before merging

guoxe (approved):

.devcontainer/Dockerfile
apljungquist: I tried reproducing the bug on ARTPEC7 because that is the armv7 SoC I have on my desk on top of c9c14d55b99bb423bcb817af78517b9cb5680b35, and it worked.
I also confirmed that the bug does not reproduce when the devcontainer is rebuilt on top of this PR.

PR #261, opened 2026-09-03
docs: Update status of vdo and larod in README

vsem-azamat:
The `vdo` crate was merged in #223 but the README still lists it as started and links to the old pull request #153. This points it to the source code and marks it Alpha, consistent with the other merged crates.

The `larod` entry now links to the current proposal #230 instead of #120.


apljungquist (approved):
Pretty uncontroversial. Apologies for not attending to this sooner. Let's see if we can get an admin approval as well so we can go ahead and merge it.

guoxe (approved):

Issue #244, opened 2026-07-11
Update `vdo`


apljungquist:
The bundled `vdo-sys` bindings have changed, but the changes are not yet accounted for in the safe `vdo` crate. Each section below corresponds to one SDK bump; when the SDK is bumped again, add a new section with the changes introduced by that bump.

## ACAP Native SDK 12.1.0 → 12.11.0 (#240)

Added:

- [ ] `VDO_WDR_MODE_AUTO` (new `VdoWdrMode` variant)
- [ ] `VDO_FORMAT_AVIF` (new `VdoFormat` variant)
- [ ] `VDO_FRAME_TYPE_AVIF` (new `VdoFrameType` variant)
- [ ] `VdoZipStreamGdr` (new type) and `vdo_zipstream_gdr_get_type`
- [ ] `VdoQuad32i`, `VdoQuad32u` (new types)
- [ ] `vdo_map_clone`
- [ ] `vdo_map_get_quad32i`, `vdo_map_set_quad32i`
- [ ] `vdo_map_get_quad32u`, `vdo_map_set_quad32u`
- [ ] `vdo_frame_get_utc_timestamp`
- [ ] `vdo_buffer_is_contiguous`
- [ ] `vdo_stream_rgb_new`
- [ ] `vdo_stream_nv12_new`
- [ ] `vdo_stream_y800_new`

Changed:

- [ ] `VdoChunk`: the `data: gpointer` field became an anonymous union with members `data: gpointer`, `chars: *mut gchar`, and `bytes: *mut guint8`

Removed:

- [ ] `vdo_channel_set_stream_profile`
- [ ] `vdo_channel_set_crop_dptz`
- [ ] `vdo_channel_apply_alpha_blending`
- [ ] `vdo_color_get_type`

PR #262, opened 2026-09-03
feat(vdo): Expose map and frame additions from SDK 12.11.0

vsem-azamat:
Part of #244.

Adds `Map: Clone`, `Map::{get,set}_quad32{i,u}` with plain `Quad32i`/`Quad32u` structs instead of the bindgen unions, `StreamBuffer::utc_timestamp` and `StreamBuffer::is_contiguous`.

The AVIF variants and `VdoWdrMode` already come through the `vdo-sys` re-exports, so they need no changes here.

Not covered: `VdoChunk`, `VdoZipStreamGdr`, the `vdo_channel_*` removals and the typed stream constructors (`vdo_stream_{rgb,nv12,y800}_new`), which this crate does not use yet. I would like to hear how you would prefer the typed constructors to fit into `StreamBuilder` before adding them.


crates/vdo/src/lib.rs
```diff
@@ -509,6 +509,15 @@ impl StreamBuffer<'_> {
         unsafe { vdo_sys::vdo_frame_get_custom_timestamp(self.raw) }
     }
 
+    /// Capture time as microseconds since the Unix epoch, from the wall clock.
+    ///
+    /// Unlike [`timestamp()`](StreamBuffer::timestamp) this can jump if the
```
apljungquist, line 514: Nice clarification! Do you remember how you learned this?
Would be awesome to document with a footnote or in the commit message so that future readers can verify it and not just take our word for it.

crates/vdo/src/lib.rs
```diff
@@ -921,6 +940,11 @@ mod tests {
             let buffer = running.next_buffer()?;
             let ts = buffer.timestamp();
             let seq = buffer.sequence_number();
+            assert!(
+                buffer.utc_timestamp() > 0,
+                "UTC timestamp should be populated"
+            );
+            assert!(buffer.is_contiguous(), "YUV frames should be contiguous");
```
apljungquist, line 947: Why do we expect these two properties to hold and how does this relate to the property ensured y the test?

crates/vdo/src/map.rs
```diff
@@ -9,7 +9,69 @@ use std::{
...
+        VdoQuad32u {
+            __bindgen_anon_1: Default::default(),
+            __bindgen_anon_2: Default::default(),
+            val: Default::default(),
+            bindgen_union_field: [self.x, self.y, self.w, self.h],
+        }
+    }
+
+    fn from_raw(raw: VdoQuad32u) -> Self {
+        let [x, y, w, h] = raw.bindgen_union_field;
```
apljungquist, line 71: How do we know that VDO always initializes all four words?

crates/vdo/src/lib.rs
```diff
@@ -1269,6 +1293,37 @@ mod tests {
...
+            x: 1,
+            y: 2,
+            w: 3,
+            h: u32::MAX,
+        };
+        map.set_quad32u(c"test_quad32u", quad_u);
+        assert_eq!(map.get_quad32u(c"test_quad32u", Quad32u::default()), quad_u);
+
+        let copy = map.clone();
+        drop(map);
```
apljungquist, line 1320: This is fine because the docs say the clone is deep and we **shouldn't** have to test the C implementation, but I'm a bit distrustful towards the docs and if we want to be sure then Claude had a good idea:

> The clone test doesn't prove the copy is deep (lib.rs, the new block at the end of the map test). It clones the map, drops the original, then reads from the copy. That would also pass if Clone only bumped the reference count (g_object_ref) instead of copying, because the copy would keep the object alive. The doc on clone() says "deep copy". To actually test that, change the original after cloning, e.g. map.set_u32(c"test_u32", 7), and check that the copy still returns 42. You can also check the other direction.

crates/vdo/src/lib.rs
```diff
@@ -527,6 +536,16 @@ impl StreamBuffer<'_> {
     pub fn is_last_buffer(&self) -> bool {
         unsafe { vdo_sys::vdo_frame_get_is_last_buffer(self.raw) != glib_sys::GFALSE }
     }
+
+    /// Returns `true` if the frame data is one contiguous region.
+    ///
+    /// Sparse buffers are spread over several chunks with gaps between them,
+    /// and only the contiguous layout is exposed by
+    /// [`as_slice()`](StreamBuffer::as_slice) and
```
apljungquist, line 544: What happens if we use `as_slice` when the buffer is not contiguous?

crates/vdo/src/map.rs
```diff
@@ -9,7 +9,69 @@ use std::{
...
+    }
+}
+
+/// Four unsigned 32-bit integers, stored in a [`Map`] under one key.
+///
+/// VDO interprets the fields either as a rectangle (`x`, `y`, `w`, `h`) or,
+/// for some keys, as a range (`min`, `target`, `max`) occupying the first
+/// three fields.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct Quad32u {
```
apljungquist, line 53: If this can be interpreted as either a rectangle or a range, maybe we should simply type it as `[u32; 4]` to not mislead users. We could add accessors to help:

```rust
/// Return the `x` value if the the value represents a rectangle
fn x(&self) -> u32 {self.0[0]}
```

Or even better, have different types for rectangles and ranges, but it's not clear to me how we would figure out what type something is or if it is even possible for us to do.

apljungquist (commented):
Nice! Some things are a bit unclear, but I think that's because of the underlying API.

What prompted you to make this change? Do you need them for anything that you are building, or just having fun contributing, or something else?

> I would like to hear how you would prefer the typed constructors to fit into StreamBuilder before adding them.

I'll try to take some time to think about this.

apljungquist:
I removed the bit about tests not having run on hardware from the PR description (commit message) after running them myself.

apljungquist (approved):

